PR #2, opened 2020-04-09
Add a license

MarkIannucci:
Hi @jmezach , this looks like something we'd really like to use for some of our internal projects at Denver Health.  Any chance I can get you to add a license?  I suggested MIT simply because it is well known and permissive which I think may be consistent with what [you wrote](https://jmezach.github.io/post/introducing-msbuild-sdk-sqlproj/).

Please feel free to squash on completion since I didn't realize that the GitHub License file would automatically include my name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -347,3 +347,6 @@ src/MSBuild.Sdk.SqlProj/tools/
 
 # Ignore Mac specific files
 .DS_Store
+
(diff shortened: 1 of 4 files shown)


jmezach:
@MarkIannucci Thanks for this. I probably should have thought about this earlier ;). I've made some additional changes to ensure that the LICENSE ends up in the NuGet package as well. I'll do a 1.0.1 release with those changes.

Issue #9, opened 2020-04-25
Post/Pre deployment scripts support


molszews:
Seems like these are not supported.

I've played a little bit with appending such scripts, seems like simple appending of the .sql to resulting dacpac (zip) should work (https://the.agilesql.club/2019/03/how-can-we-merge-multiple-dacpacs-into-one/), however, I'm not sure how to tackle sqlcmd syntax support.
Running `sqlcmd -i postdeployment_entryfile.sql -e -r >postdeploy.sql` generates correct file, but bundling sqlcmd within the target project may be difficult?

ErikEJ:
Any thoughts about the design of this?

``` xml
  <ItemGroup>
    <PostDeploy Include="Postdeployment\Script.PostDeployment.sql" />
  </ItemGroup>

```
If this is passed to the command line tool, maybe we can just expect sqlcmd to be present, and run it to create the consolidated script(s), and add it like this: https://github.com/GoEddie/DacpacMerge/blob/master/src/MergeEm/Program.cs#L118

molszews:
well, using sort of naming convention won't be enough? there can be only one post/pre deploy script, we can simply assume it should be available under `Post-deployment\postdeploy.sql`

ErikEJ:
You can link multiple script with r: sqlcmd syntax - and we do.

molszews:
yeah, but you specify only the entry point inside the .csproj, after that you can link every other file?
btw, sqlcmd could be embedded in a way this package embeds chromedriver binary https://github.com/jsakamoto/nupkg-selenium-webdriver-chromedriver/#where-is-chromedriverexe-saved-to

ErikEJ:
But for v1, we could of course just go by convention, and expect the script to be merged.

ErikEJ:
Not sure sqlcmd is simply a single .exe..

jmezach:
I wonder how SSDT treats those scripts. I'm guessing it may just be added to the model as is and the passed to sqlcmd at deployment time. Does anyone have an example `.dacpac` that is build by SSDT which includes a pre- and/or postdeployment script that they are willing to share so I can have a look?

molszews:
Resulting .dacpac archive contains a single .sql file which looks like the result of sqlcmd command I've posted before. It looks like there is nothing in the model related to that postdeploy.sql, it is simply appended to the zip archive. I will try to provide some samples

ErikEJ:
[...]

PR #13, opened 2020-05-05
Initial steps toward #9

ErikEJ:


src/BuildDacpac/PackageBuilder.cs
```diff
@@ -55,6 +57,31 @@ public void AddInputFile(FileInfo inputFile)
             Model.AddObjects(File.ReadAllText(inputFile.FullName));
         }
 
+        public void AddPrePostScripts(FileInfo pre, FileInfo post, FileInfo outputFile)
```
jmezach, line 60: I think that for maintainability it might be better to have separate `AddPreDeploymentScript` and `AddPostDeploymentScript` methods here, which delegate to a private `AddDeploymentScript` method. That would lower the number of tests as well, since there are less branches to test. Then we would just call the appropriate method(s) depending on the passed arguments.

jmezach:
@ErikEJ Thanks for this, this is looking great. I made one comment, but I think this is pretty much good to go. Of course this doesn't have support for includes yet, as discussed in #9. I did some experiments today and I think we can even make that work without re-inventing the wheel, but that will need some more experimenting.

Do you feel confident to do the MSBuild stuff yourself this time, or would you like me to do it?

ErikEJ:
PR updated - and please go ahead with the MsBuild changes - I am watching and learning!

jmezach:
This looks great, thanks. I'll go ahead and merge this and I hope to get to the MSBuild stuff sometime this week.

PR #29, opened 2020-07-30
Replace backslashes in references dacpac paths with forward slashes

### LINKED ISSUE #28: Referencing other dacpacs fails on macOS with 'Unable to find reference file' ArgumentException
I'm working on a POC for adopting this project to manage my organization's database builds to make them cross-platform. Our builds typically combine 3-5 dacpacs into a composite database project. When attempting to build a composite project that includes a `PackageReference` on a Mac, I received an error that the SDK was 'Unable to find reference file'.

Here's my .csproj file (unnecessary properties removed):

``` xml
<Project Sdk="MSBuild.Sdk.SqlProj/1.1.0">

  <!-- Database configuration properties -->
  <PropertyGroup>
    <!-- ... -->
  </PropertyGroup>

  <!-- File includes / excludes -->
  <!-- MSBuild.Sdk.SqlProj Sdk defaults to including all *.sql files -->
  <ItemGroup>
    <Content Remove="Scripts/**/*.sql" />
    <Content Remove="target/**/*.sql" />
    <PostDeploy Include="Scripts/Script.PostDeployment.sql" />
    <PreDeploy Include="Scripts/Script.PreDeployment.sql" />
  </ItemGroup>

  <!-- References to dacpac dependencies -->
  <ItemGroup>
    <PackageReference Include="Base" Version="7.6.0" />
  </ItemGroup>
</Project>
```

Here's the error I receive on performing a build or pack:

```
Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
   ---> System.ArgumentException: Unable to find reference file /Users/jrosenberg/.nuget/packages/Base/7.6.0\tools\Base.dacpac (Parameter 'referenceFile')
     at MSBuild.Sdk.SqlProj.BuildDacpac.PackageBuilder.AddReference(FileInfo referenceFile) in /home/runner/work/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj/src/BuildDacpac/PackageBuilder.cs:line 32
     at MSBuild.Sdk.SqlProj.BuildDacpac.Program.BuildDacpac(PackageBuilderOptions options) in /home/runner/work/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj/src/BuildDacpac/Program.cs:line 57
     --- End of inner exception stack trace ---
     at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
     at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
     at System.Delegate.DynamicInvokeImpl(Object[] args)
     at System.Delegate.DynamicInvoke(Object[] args)
     at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
     at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseParseErrorReporting>b__20_0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass15_0.<<UseHelp>b__0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass22_0.<<UseTypoCorrections>b__0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseSuggestDirective>b__21_0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseParseDirective>b__19_0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseDebugDirective>b__11_0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<RegisterWithDotnetSuggest>b__10_0>d.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass13_0.<<UseExceptionHandler>b__0>d.MoveNext()
```

I'm pretty sure the issue is the backslashes in the path: '/Users/jrosenberg/.nuget/packages/Base/7.6.0\tools\Base.dacpac'. I notice that the Sdk.targets folder currently has a mix of forward- and backslashes. I've been able to fix this locally; I will send a pull request shortly.

jeffrosenberg:
Resolves #28. I don't currently have a Windows machine available to test on, but I believe forward slashes will work on Windows. I also see other paths in Sdk.targets that use forward slashes, such as `<BuildDacpacExe>$(MSBuildThisFileDirectory)../tools/$(BuildDacpacTfm)/BuildDacpac.dll</BuildDacpacExe>`.

There might be better solutions for this, but it solved my immediate problem, so thought I'd send the PR. Hope it helps!


ErikEJ:
Thanks, Maybe it would be better to use GetFullPath in the DacpacBuilder code?

jmezach:
I agree with @ErikEJ.  I think MSBuild doesn't really care about the difference between both path separators and automatically translates between them. The real problem is probably in the BuildDacpac tool. It should probably normalise the path to whatever is relevant on the current platform.

jeffrosenberg:
Sure @ErikEJ @jmezach, I can take a look at that. I have to turn my attention to a few other things at work, but I'll try to update the PR today or tomorrow.

ErikEJ:
Something like this:

```
        public void AddReference(FileInfo referenceFile)
        {
            // Ensure that the model has been created
            EnsureModelCreated();

            // Make sure the file exists
           var fileName = Path.GetFullPath(referenceFile.FullName);
            if (!File.Exists(fileName))
            {
                throw new ArgumentException($"Unable to find reference file {fileName}", nameof(referenceFile));
            }

            Console.WriteLine($"Adding reference to {fileName}");
            Model.AddReference(FileName);
        }

```
Do not have a Mac handy for testing....

jeffrosenberg:
@ErikEJ I was able to try that on my Mac, and it didn't work -- it still spits out a path with mixed backslashes and forward slashes, and I still get the original error. 

I'm looking at whether there's another way to standardize the path, but I think the problem is that the path being passed into the `AddReference()` method is already a `FileInfo` object with mixed directory separators; it seems that `Path.GetFullPath` doesn't know how to handle that, at least on a Mac.

ErikEJ:
Looks like GetFullPath does not touch the separators.

jeffrosenberg:
We could replace the separators in code, but at that point it seems to me that it's basically the same thing as replacing them in `Sdk.targets`, since any paths other than those hardcoded there will have the correct separators for the user's OS. 

I'd note that [Microsoft's documentation on path separators](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=netcore-3.1) recommends just using forward slashes everywhere:

> When you are using .NET Core to develop applications that run on multiple platforms:
> 
> - If you prefer to hard-code the directory separator character, you should use the forward slash (/) character. It is the only recognized directory separator character on Unix systems, as the output from the example shows, and is the AltDirectorySeparatorChar on Windows.

There are other options mentioned on that page, but since `Sdk.targets` hardcodes certain relative paths, they don't seem applicable to me.

jmezach:
Yeah, I think so too. What I'd like to do though is to expand the test set so that we cover the scenario that went wrong here to make sure things don't regress later on. There's a test project in `test/TestProjectWithSDKRef` which, as the name suggests, uses the `<Project Sdk="MSBuild.Sdk.SqlProj/<version>">` syntax to simulate the end user scenario as closely as possible. It restores that package from a local folder in which the NuGet package gets downloaded that was build in the previous stage.

I guess that if during the build stage we'd also produce a NuGet package based on this SDK, download that as well during the test stages (these run on multiple platforms) and then reference that NuGet package from the `TestProjectWithSDKRef` we can test this scenario.

jmezach:
@jeffrosenberg Are you comfortable adding a test scenario for this, or do you need some help?

jeffrosenberg:
@jmezach I'm not sure whether I follow, but I'll give it a try tonight and let you know if I'm not able to figure it out. I'll do my best to have an update to the PR ready for tomorrow.

jeffrosenberg:
@jmezach my apologies, but after looking through the code and build pipeline, I'm not following. You said:

> I guess that if during the build stage we'd also produce a NuGet package based on this SDK, download that as well during the test stages (these run on multiple platforms) and then reference that NuGet package from the TestProjectWithSDKRef we can test this scenario.

But it seems to me that's already what the build does? `main.yml` already does a build, pack, and upload of MSBuild.Sdk.SqlProj, and then it looks like that artifact is downloaded against and used as the source for the build of TestProjectWithSDKRef. 

Can you maybe explain in another way what needs to be added to this? Thanks!

jmezach:
Sorry for not being clear. What I meant was that we need to have a NuGet package that is the output of a project based on this SDK, so that we can add a `PackageReference` to it from the test project. If for example we we're to create a project using the SDK with a simple table and do a `dotnet pack` on it in the build stage and upload the resulting nupkg, we can download and reference it in the test project. Does that make sense?

jeffrosenberg:
Oh, I see. So you're asking to make TestProjectWithSDKRef a composite project with a PackageReference, similar to TestProjectWithPackageReference? So we'd pack and upload a simple project -- could maybe just use TestProject, but I'd need to play around with that -- and then reference it from TestProjectWithSDKRef?

jmezach:
Yeah, that sounds good. To make it even better I think it would be good to have TestProjectWithSDKRef use something from TestProject (like a table or a function or something) so that if the PackageReference somehow doesn't work correctly building the TestProjectWithSDKRef should fail.

jeffrosenberg:
Yeah, agreed. I get it now, I'll get that change made!

jeffrosenberg:
Well, it's almost all working, except for the important part! The composite project isn't actually building; dbo.MyTable isn't being included in the project. At least we know the test works! I'll try to figure out why and get it resolved in the morning.

jmezach:
Looks like it works on Windows and Mac, but doesn't work on Linux. Looks like the reference argument isn't passed to the BuildDacpac tool from the targets file at all on Linux, while it is on Mac and Windows. Also interesting that it comes out as a warning and not as an error, but that is existing behaviour I guess.

jeffrosenberg:
Aha, got it. Apparently on deployment using sqlpackage, we now need to specify `IncludeCompositeObjects=true`. Added that property and the deployment succeeds now. I confirmed locally that two tables are created, and both populated successfully by a post-deploy script. I added a blurb about that to the readme.

jmezach:
@jeffrosenberg Unfortunately I still see warnings in the build on Linux, but not on Windows and Mac. See the [results](https://github.com/jmezach/MSBuild.Sdk.SqlProj/pull/29/checks?check_run_id=945026450#step:9:44). Looks like the reference isn't being added on that platform. Do you mind looking into what happens there?

jeffrosenberg:
Oh you're right, I see it. The step was reporting green so I didn't notice. Yes, I'll look at that later today. My first inclination is that maybe it's an issue with case sensitivity? I'll spin up a linux container and try it out.

jeffrosenberg:
@jmezach I spent some time trying to work this out, but haven't been able to nail it down. It looks to me like the issue stems from the `Condition="Exists('%(PhysicalLocation)/tools/%(Identity).dacpac')` in `Sdk.targets`. But I'm not certain whether that's due to a case-sensitivity issue on Linux or some other problem. I'm also not sure whether this is a new issue created by my first commit, or whether it just hadn't been caught because the test phase previously didn't include a dacpac reference. I'm guessing the latter, but haven't tested it yet.

I'm happy to keep looking into this if you'd like, but if you think you could solve this more easily, that's fine with me too. If you'd like me to keep looking at it, I might open up a separate draft PR to keep testing this using the Github workflow, without spamming you with notifications on failed builds.

jmezach:
I've pushed a change to the CI pipeline that will hopefully give us an MSBuild binary log that we can analyse so that we can figure out what's going on. It all looks good to me so far, so I'm not entirely sure what's going on either. I hope the binary log can tell us a little bit more.

jmezach:
Right, this is getting weird. I had a look at the binary log and it seems that it is resolving the package reference `TestProject` to the correct physical location as indicated by the message `Resolved package reference TestProject to /home/runner/.nuget/packages/TestProject/1.2.0-beta.7.geb09cfd0cf` that is written by [this task](https://github.com/jmezach/MSBuild.Sdk.SqlProj/blob/e803b231ca3e13a3ddfefef93c44ebeb182ec0b9/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L122) from `Sdk.targets`.

However the `ItemGroup`defined [here](https://github.com/jmezach/MSBuild.Sdk.SqlProj/blob/e803b231ca3e13a3ddfefef93c44ebeb182ec0b9/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L119) checks for the existence of a `.dacpac` with the same name as the package (`TestProject` in this case) inside the `tools` folder.  Looking at the NuGet package I would assume that such a file should be there. But it looks like the `Exists()` check there just returns `false` for some reason. Maybe the `dotnet restore` on Linux is somehow restoring every file within the NuGet package lowercase or something?

jeffrosenberg:
@jmezach my observation has been that NuGet seems to restore everything as lowercase, at least on my Mac. But the Mac file system is case-insensitive, so it's not an issue there. So yes, it's my hunch that's what's going on here. But it seems wild to me that an issue regarding MSBuild `Exists` conditions wouldn't have come up and been addressed by now, and this definitely seems to be happening at the MSBuild level, not within the `PackageBuilder` class.

jmezach:
Hmm, yeah, you're right. Maybe we should just lower case both the `%(PhysicalLocation)` and `%(Identity)` in that condition and see if that resolves the issue. I think we can use MSBuild's Property Functions to do that using `string.ToLower()`.

ErikEJ:
Be careful with ToLower http://csharp.tips/tip/article/401-turkish-i-in-csharp

jmezach:
Yeah, I'm aware of the issues with ToLower, but I don't think we have much of a choice if NuGet is laying out files in lower case. I'll see if I can make this patch when I have some time.

jeffrosenberg:
@jmezach I can add that change to this PR, unless you'd like to take it from here?

jmezach:
I did have a look at this, but couldn't find a quick way to implement this. But if you have an idea, go for it :).

jeffrosenberg:
@jmezach I believe I still see the same issue in the builds

global.json
```diff
@@ -0,0 +1,5 @@
+{
+    "sdk": {
+        "version": "3.1.301"
+    }
+}
```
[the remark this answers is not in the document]
jmezach: No, that wasn't my intent. I'm running .NET 5 previews locally and things weren't working, so I added this and then accidentally committed it. Got it fixed already. Guess we'll need to figure out why it's not working with .NET 5 SDK's though.

jmezach:
I had a look at the binary log again, and I'm now seeing what I was expecting:

`Message: Resolved dacpac file from package TestProject to /home/runner/.nuget/packages/testproject/1.2.0-beta.10.gbe99a3893b/tools/testproject.dacpac`

But apparently that still doesn't pass the `Exists()` check. Looking at my local NuGet package cache (on a Mac) it looks like while the folder containing the package is lowercased, the contents of the package are as they are packaged. So I guess we need to use the lower case identity after the `$(NuGetPackageRoot)` but then use the regular cased identity for the `.dacpac` filename.

jeffrosenberg:
Okay, I think we're finally there! I'm not seeing the issue in the Linux build anymore, and macOS and Windows also look good. One last thing -- I changed the message priority for the dacpac resolutions from "Normal" to "High" to help with debugging, since I'm having trouble reading the binary logs. Do you want me to set them back? Also, am I good to rebase this into a single commit?

jmezach:
Yep, it looks to be working now. I'll go over the PR again to make sure it's in tip-top shape now. And yeah, we'll probably want to lower the importance of the messages back to normal again to avoid spamming the user with all kinds of diagnostic output.

As for the binary logs, I think there's currently only a Windows based application that can read them. I have an Azure VM with Windows setup that I use to read those logs, since I'm no a Mac, but it would be nice if such a tool was available on Mac as well. I don't think there are any plans for that at the moment though.

.github/workflows/main.yml
```diff
@@ -51,16 +51,27 @@ jobs:
...
+    
+    # Package TestProject
+    - name: dotnet pack TestProject
+      run: dotnet pack -c Release ./test/TestProject/TestProject.csproj
+
+    # Upload TestProject package
+    - name: upload TestProject
+      uses: actions/upload-artifact@v1
+      with:
+        name: nuget-package
```
jmezach, line 73: Nit: We might want to rename this artifact to nuget-packages since it now contains multiple packages.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -112,15 +112,19 @@
     <ItemGroup> 
       <!-- Resolve all package references to their physical location first -->
       <_ResolvedPackageReference Include="%(PackageReference.Identity)">
-        <PhysicalLocation>$(NuGetPackageRoot)%(PackageReference.Identity)\%(PackageReference.Version)</PhysicalLocation>
+        <PhysicalLocation>$(NuGetPackageRoot)%(PackageReference.Identity)/%(PackageReference.Version)</PhysicalLocation>
       </_ResolvedPackageReference>
+      <_ResolvedDacpacFile Include="@(_ResolvedPackageReference)">
+        <DacpacFile>$([System.String]::new('%(PhysicalLocation)/tools/').ToLower())%(Identity).dacpac</DacpacFile>
+      </_ResolvedDacpacFile>
```
jmezach, line 119: I think we can actually simplify this back to a single ItemGroup again. Something like this should work:

```
$(NuGetPackageRoot)$([System.String]::new('%(PackageReference.Identity)/%(PackageReference.Version)').ToLower())/tools/%(Identity).dacpac
```

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -112,15 +112,19 @@
...
-      <DacpacReference Include="@(_ResolvedPackageReference)" Condition="Exists('%(PhysicalLocation)\tools\%(Identity).dacpac')" />
+      <DacpacReference Include="@(_ResolvedDacpacFile)" Condition="Exists(%(DacpacFile))" />
     </ItemGroup>
-  
-    <Message Importance="normal" Text="Resolved package reference %(_ResolvedPackageReference.Identity) to %(_ResolvedPackageReference.PhysicalLocation)" />
-    <Message Importance="normal" Text="Resolved database package references: @(DacpacReference)" />
+
+    <Message Importance="High" Text="Resolved package reference %(_ResolvedPackageReference.Identity) to %(_ResolvedPackageReference.PhysicalLocation)" />
+    <Message Importance="High" Text="Resolved dacpac file from package %(_ResolvedDacpacFile.Identity) to %(_ResolvedDacpacFile.DacpacFile)" />
+    <Message Importance="High" Text="Resolved database package references: @(DacpacReference)" />
```
jmezach, line 127: As mentioned we'll probably want to lower the `Importance` back to `normal` again here.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
```diff
@@ -1,5 +1,15 @@
 <Project Sdk="MSBuild.Sdk.SqlProj/#{NBGV_NuGetPackageVersion}#">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
+        <SqlServerVersion>Sql110</SqlServerVersion>
```
jmezach, line 4: That's interesting. Was this change necessary to make it work? I think there's a default somewhere.

global.json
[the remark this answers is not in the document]
jmezach: I've filed #30 to look at this issue.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
jeffrosenberg: It's one of the changes I made when I was trying to get the composite project to build. I know I've had issues with that in the past -- build issues when the projects aren't on the same version -- but I'm not sure whether this was actually part of what fixed the problem. I'll experiment with removing it again and see if anything breaks.

.github/workflows/main.yml
jeffrosenberg: Maybe I'm misunderstanding, but I don't believe it does. `TestProjectWithSDKRef` now contains multiple packages, but `TestProject` is still just the one: `TestProject/bin/release/netstandard2.0/TestProject.version.nupkg`.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: Done. My version is slightly different, as I believe we want `$NuGetPackageRoot)` inside the string function.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: Done.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
jeffrosenberg: Looks like it wasn't necessary, I've backed it out.

.github/workflows/main.yml
jmezach: You're right, but the MSBuild.Sdk.SqlProj package is also uploaded using the same name a little bit further up in the workflow file. I guess GitHub actions combines both packages into the same artifact (somehow), which is fine, but we should then at least make it plural to indicate what it is ;).

jeffrosenberg: Oh, I see. Okay, no problem.

jmezach:
Cool! I think I will squash merge this to `master` and then cherry-pick it to the release branch for 1.1.0. That should trigger a release of 1.1.1.

jeffrosenberg:
Great! Thanks for this project, I think it will be a big help to my organization! As I'm working through our implementation, I may submit some additional PRs if we happen to run into any other issues.

ErikEJ:
Thanks Jeff, great improvement to the project!

jmezach:
Yes, thanks for this Jeff. Really appreciate you taking the time to work with us on this. And feel free to send more PR's in the future. They are always welcome :).

Issue #17, opened 2020-05-06
Occasional issue during CI


jmezach:
It sometimes happens that CI fails due to SqlPackage trying to deploy the .dacpac to the container, but the container hasn't yet fully started. I believe there is a way to wait for the container to start before the job runs, but I'll need to dive in to see how that works exactly.

jmezach:
Fixed by #33

Issue #32, opened 2020-08-07
Proposal: Adding support for deploying


jmezach:
I've been considering adding support for deploying a project that uses this SDK to a SQL Server as part of a `dotnet publish`. I have experimented with this and it is fairly easy to extend the publishing pipeline by adding a `Publish` target to the `Sdk.targets` file and do something from there. But before I go and implement this I was wondering if this would be considered useful for other folks using this package.

I do see a couple of challenges. First of all, we will need to have `sqlpackage.exe` to do the actual publishing. On Windows you get this as part of the SSDT tools, but on Linux and Mac there isn't an easy way to acquire it. You would need to download it and then make it available on the path for this package to use it. Or we could include it in this package, but I'm not sure if we're even allowed to redistribute it. Ideally `sqlpackage.exe` would be available as a .NET Core Global Tool so we could run `dotnet tool install SqlPackage` but that is not how it has been implemented today. Maybe @dzsquared knows if this is changing at some point?

Of course we'd also need to know the server and database name to deploy to. We could probably use localhost and the name of the project as sane defaults, but allow these to be overridden through properties in the project file.

So basically I'm looking for some feedback on this idea. @ErikEJ What do you think?

ErikEJ:
Why do we need sqlpackage(.exe) ? 

https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.dac.dacservices.deploy?view=sql-dacfx-150 

(Not claiming it will be simple to support all options)

jmezach:
Hmm, yeah, you're right, we don't need to use SqlPackage.exe to deploy it, we could just build a tool ourselves that does it and then run that. Would be quite a bit of work to support all options yes, but maybe we don't need to support them all since it will be mostly used for local development I think.

Maybe we need to rename the current BuildDacpac tool and add the support there, or have a separate tool. They have a different purpose of course, but maybe including yet another tool in the same package is a bit too much overhead. We would be duplicating dependencies in that case.

jmezach:
This has been implemented in #33 and should be released soon as 1.2.0. I'll leave this open until that version is out.

jmezach:
[1.2.0](https://www.nuget.org/packages/MSBuild.Sdk.SqlProj/1.2.0) is out on NuGet.org which has support for this. Thanks @ErikEJ for reviewing the change.

PR #33, opened 2020-08-07
Add support for publishing a project to a SQL Server

jmezach:
As discussed in #32 I've added initial support for deploying a project to a SQL Server as part of a `dotnet publish` command. It's a fairly large change, mostly due to the rename of the `BuildDacpac` project to `DacpacTool` to more accurately describe what it does.

I still need to write some tests, but the preliminary results on my local machine are promising. I've successfully published a project both on Windows and on Mac. We'll also want to add something to the CI pipeline to make sure we test this functionality there.

As with the model options we can control the various aspects of the deployment through MSBuild properties. There are a couple of them that will need some additional work still because they aren't simple values.

I've also dropped support for .NET Core 2.2, since that version is no longer being supported by Microsoft.


ErikEJ:
Looks deceptively simple!

How would the publish command look?

jmezach:
On Windows I can get by with just `dotnet publish` because of Windows authentication. That obviously doesn't work on Mac and Linux so you'll need to specify a username (using `/p:TargetUserName=<username>`) and it will then prompt you for a password. We might want to add support for specifying the password on the command line though for CI scenarios.

ErikEJ:
Very cool!

jmezach:
I think I'm about done here. I'll need to write some docs on how to use this and add this to the CI pipeline to make sure it gets properly tester, but pretty much everything I need is working right now. @ErikEJ Do you mind giving this a review?

ErikEJ:
I will be happy to, just waiting for the temperature to get lower so I can think straight 😄 (32 degrees Celsius)

jmezach:
Haha, fair enough. Yeah, it's pretty hot here right now as well.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
```diff
@@ -115,5 +115,101 @@
...
+      TreatVerificationErrorsAsWarnings;
+      UnmodifiableObjectWarnings;
+      VerifyCollationCompatibility;
+      VerifyDeployment;
+    </KnownDeployProperties>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Define some useful defaults for deployment -->
+    <TargetServerName>localhost</TargetServerName>
```
ErikEJ, line 211: Should this not be: (local)

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -145,16 +145,37 @@
...
+    <Message Importance="Low" Text="Running command: $(DacpacToolCommand)" />
+    <Exec Command="$(DacpacToolCommand)"
           Condition="'$(DesignTimeBuild)' != 'true' AND '$(BuildingProject)' == 'true'"/>
 
     <!-- Copy any referenced .dacpac packages to the output folder -->
     <Copy SourceFiles="@(IncludedDacpacReferenceFiles)" DestinationFolder="$(TargetDir)" />
   </Target>
+
+  <!-- Performs a publish of the .dacpac produced by build to a configurable target server -->
+  <Target Name="Publish" DependsOnTargets="Build">
```
ErikEJ, line 160: Does this mean you can simply run "dotnet publish" and "dotnet build" will then run first?

ErikEJ:
LGTM, just a couple of comments/questions.

.github/workflows/main.yml
```diff
@@ -150,6 +154,12 @@ jobs:
           SA_PASSWORD: JdMsKZPBBA8kVFXVrj8d
         ports:
           - 1433:1433
+        options: >-
```
ErikEJ, line 157: Does this fix #17 ?

jmezach: I think it does. I haven't seen that issue again after this change, but then again it didn't occur every time. But I think I'll close #17 when this is merged and if it occurs again we'll open a new issue.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jmezach: Yes. We need to have a .dacpac build before we're able to deploy it. I did notice that the build is always performed regardless of whether anything has actually changed so I filed #34 to fix that.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
jmezach: Hmm, not sure. I don't know if that works cross-platform while I'm fairly sure that localhost works everywhere. Then again, I think `(local)` is more idiomatic SQL so I'll test to see if it works cross-platform.

jmezach: Looks like that works on Mac as well. I only needed to add some quotes around the server name, but I guess that's a good thing to do anyway.

jmezach:
I've addressed the comments. I want to add one more thing, which is to allow projects or other NuGet packages to add their own thing before or after the publish step which is a common thing to do. Then I'll write up some docs in the README for this feature and I think we can get this merged to master. I'll then give the beta a try on a real project to make sure everything is good and then I think we can release this as 1.2.0.

ErikEJ:
Fantastic! We are years ahead of Microsoft with this!

jmezach:
Looks like I don't need to do anything special to allow the project or any other NuGet package to extend the publishing pipeline. Simply adding a `<Target Name="DoSomething" AfterTargets="Publish">` is enough. Better get on with writing those docs then :).

jmezach:
@ErikEJ I've finished writing the docs. Do you mind proofreading the README.md to make sure I didn't miss anything?

ErikEJ:
:shipit:

jmezach:
I've made the suggested changes so I think we're ready to merge this when the CI finishes ;).

jmezach:
Wait a minute. I just remembered that I wanted to have a dotnet publish in the CI pipeline so that we can see that working. Let me just add that before we merge this ;).

jmezach:
Right, I got that fixed. We might want to eventually check whether the right objects are created in the database after deployment (for both `SqlPackage` and `dotnet publish`) but for now I'll settle for seeing it deploy correctly. Let's get this merged into `master` and I'll give the beta a whirl on a real project.

jmezach:
Works like a charm. Let's get this released!

AraHaan:
I noticed this sdk and I got to say that it looks so far good. However msbuild seems to lack an ``PublishOnBuild`` property that I know of.

Could this project perhaps do something like this to the sdk's targets file?

```xml
<Project>
  <!-- stuff currently in the Sdk.targets file. -->

  <Target Name="PublishSqlProject" AfterTargets="Build" Condition="$(PublishOnBuild)">
      <MSBuild Projects="$(MSBuildProjectFullPath)"
           Targets="Publish"
           BuildInParallel="true"
           Properties="TargetFramework=$(TargetFramework)" />
  </Target>

</Project>
```

ErikEJ:
Build on Publish would change the schema of a database on each build. Why would that be useful?

AraHaan:
Some people build in debug to test the thing to make sure it builds, however for them in Release mode they typically might want it to auto-publish too.

ErikEJ:
In my opinion, only sqlpackage or the DaxFX api should be used to deploy a dacpac. Not this SDK.

jeffrosenberg:
I can see an argument for publish on build for a local dev database, particularly something like a `LocalDb` instance that's being used to debug Entity Framework code. Honestly, I feel like we should either (1) support publish on build if the PR looks good (#286), or (2) stop supporting publish entirely, since we've consistently tried to steer people away from it for non-trivial use cases. 

I guess the question here is whether we still feel like supporting publishing solves a problem? The build side of this project meets a clear need, in that there still isn't a good way to build a database project on non-Windows systems. Not so for publishing, so maybe it would make sense to focus on our "core competency" and simply direct people to use SqlPackage as a post-build step for use cases like this one.

jmezach:
We are using `dotnet publish` a lot internally for setting up local databases that are then used for running tests and it works just fine for that scenario. We're even using it in CI pipelines now for deploying the database to a (typically containerized) SQL Server instance so that we can run tests against it. I think `dotnet publish` works fine for those scenario's where you want a brand new database setup based on your project.

But for deploying to any other environment where you have to deal with an existing database I feel like people should use `SqlPackage.exe` instead as it allows far much control over the deployment process.

Issue #36, opened 2020-08-17
PackageReferences don't resolve wildcard versions


jeffrosenberg:
When using the `<PackageReference>` element to build a composite database project, the referenced version should support the use of "floating versions", i.e. wildcards, per the [documentation here](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#floating-versions). When using this SDK (and adding a few debugging messages to see what's going on behind the scenes), it appears that it is not resolving the wildcard to the most recent version, but just using the wildcard literally. 

The following is a simplified example of two .csproj files I'm building into a composite database:

### Base.csproj ###
``` xml
<Project Sdk="MSBuild.Sdk.SqlProj/1.1.1">

  <!-- Database configuration properties -->
  <PropertyGroup>
    <!-- ... -->
  </PropertyGroup>

  <!-- File includes / excludes -->
  <!-- MSBuild.Sdk.SqlProj Sdk defaults to including all *.sql files -->
  <ItemGroup>
    <Content Remove="target/**/*.sql" />
  </ItemGroup>
```

### Plus.csproj ###
``` xml
<Project Sdk="MSBuild.Sdk.SqlProj/1.1.1">

  <!-- Database configuration properties -->
  <PropertyGroup>
    <!-- ... -->
  </PropertyGroup>

  <!-- File includes / excludes -->
  <!-- MSBuild.Sdk.SqlProj Sdk defaults to including all *.sql files -->
  <ItemGroup>
    <Content Remove="Scripts/**/*.sql" />
    <Content Remove="target/**/*.sql" />
    <PostDeploy Include="Scripts/Script.PostDeployment.sql" />
    <PreDeploy Include="Scripts/Script.PreDeployment.sql" />
  </ItemGroup>
  <!-- References to dacpac dependencies -->
  <ItemGroup>
    <PackageReference Include="Base" Version="7.6.*" />
  </ItemGroup>
</Project>
```

When I try to build the "Plus" package, I see the following messages from Sdk.targets:

```
  Determining projects to restore...
  All projects are up-to-date for restore.
  PackageReference.Identity is Base
  Resolved dacpac file from package Base to /users/jrosenberg/.nuget/packages/base/7.6.*/tools/Base.dacpac
  Resolved database package references: 
```

Interestingly, if I change the dependency version to 7.7.*, MSBuild stops me before even calling this SDK, so MSBuild and NuGet appear to be resolving the dependency version.  

```
[...]

PR #37, opened 2020-08-19
Use the MSBuild Pkg<package-id> property for package location (#36)

jeffrosenberg:
For packages that meet certain criteria (such as having a tools folder), MSBuild will automatically include a property that resolves their location. This is more accurate because it handles floating versions$

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -112,9 +112,9 @@
     <ItemGroup> 
       <!-- Resolve all package references to their physical location first -->
       <_ResolvedPackageReference Include="%(PackageReference.Identity)">
-        <!-- Determine technical name of package (ie. DeVries.WMS.Database -> DeVries_WMS_Database) -->
+        <!-- Determine technical name of package (ie. Foo.Bar.Database -> Foo_Bar_Database) -->
         <PackageName>@(PackageReference->'%(Identity)'->Replace('.', '_'))</PackageName>
-        <!-- Prepend Pkg to technical name from above and resolve variable (ie. PkgDeVries_WMS_Database -> %home%\.nuget\packages\devries.wms.database\<version> -->
+        <!-- Prepend Pkg to technical name from above and resolve variable (ie. Foo_Bar_Database -> %home%\.nuget\packages\foo.bar.database\<version> -->
         <PhysicalLocation>$(Pkg%(_ResolvedPackageReference.PackageName))</PhysicalLocation>
         <!-- 
           If no Pkg<package-id> property is available, fall back to deriving the physical location from several other properties.
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -112,7 +112,17 @@
     <ItemGroup> 
       <!-- Resolve all package references to their physical location first -->
       <_ResolvedPackageReference Include="%(PackageReference.Identity)">
-        <DacpacFile>$([System.String]::new('$(NuGetPackageRoot)%(PackageReference.Identity)/%(PackageReference.Version)').ToLower())/tools/%(Identity).dacpac</DacpacFile>
+        <!-- Determine technical name of package (ie. DeVries.WMS.Database -> DeVries_WMS_Database) -->
+        <PackageName>@(PackageReference->'%(Identity)'->Replace('.', '_'))</PackageName>
+        <!-- Prepend Pkg to technical name from above and resolve variable (ie. PkgDeVries_WMS_Database -> %home%\.nuget\packages\devries.wms.database\<version> -->
+        <PhysicalLocation>$(Pkg%(_ResolvedPackageReference.PackageName))</PhysicalLocation>
+        <!-- 
+          If no Pkg<package-id> property is available, fall back to deriving the physical location from several other properties.
+          This isn't guaranteed to be correct, particularly when floating versions are used, but will successfully resolve most of the time.
+        -->
+        <PhysicalLocation Condition="'%(_ResolvedPackageReference.PhysicalLocation)'==''">$([System.String]::new('$(NuGetPackageRoot)%(PackageReference.Identity)/%(PackageReference.Version)').ToLower())</PhysicalLocation>
+
+        <DacpacFile>%(_ResolvedPackageReference.PhysicalLocation)/tools/%(Identity).dacpac</DacpacFile>
       </_ResolvedPackageReference>
     
       <!-- Build a list of package references that include a dacpac file matching the package identity in their tools folder -->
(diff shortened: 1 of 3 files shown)


jeffrosenberg:
Looks like this fails on dotnet 2.1, with the error `error : '1.*-*' is not a valid version string.` That makes sense, as it's a relatively recent change. I'll need to give some thought to the best way to clean up the tests. I think it would be valid to test the floating versions differently on 2.1 and 3.1, but I don't want to complicate the tests too much.

jmezach:
Yeah, that makes sense. I guess we can make the `PackageReference` conditional based on some property which we pass in on the command line. Then we can have the 3.1 jobs pass that property, while we don't pass it for the 2.1 jobs.

jeffrosenberg:
@jmezach makes sense to me. I have a thought on how to do that, but I can't test it locally so apologies for what's likely to be a few rounds of pushing changes and running tests.

jmezach:
You can probably test it locally by having the latest 2.1 SDK installed and putting a global.json somewhere to force that version to be used. Not that I mind testing it through CI though, but I think it's always easier to test locally first ;).

jeffrosenberg:
The issue I had is testing some of the conditionals in the Github actions file. What I've done here is to expand the use of `cschleiden/replace-tokens` in a few ways:

* Add a new `#{DependencyVersion}#` placeholder
* For dotnet 3.1, replace `#{DependencyVersion}#` with a double wildcard: `1.*-*`
* For dotnet 2.1, take a two-step approach:
  1. First replace `#{DependencyVersion}#` with another token expression: `#{NBGV_Version}#-*`
  2. Then replace the `#{NBGV_Version}#` token in the next step to resolve the major/minor/patch version

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -112,7 +112,17 @@
     <ItemGroup> 
       <!-- Resolve all package references to their physical location first -->
       <_ResolvedPackageReference Include="%(PackageReference.Identity)">
-        <DacpacFile>$([System.String]::new('$(NuGetPackageRoot)%(PackageReference.Identity)/%(PackageReference.Version)').ToLower())/tools/%(Identity).dacpac</DacpacFile>
+        <!-- Determine technical name of package (ie. DeVries.WMS.Database -> DeVries_WMS_Database) -->
```
jmezach, line 115: I'm guessing you got this comment from my original commit since this is referring to an internal package here. We should probably replace `DeVries.WMS.Database` here with something like `Foo.Bar.Database` or something.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
```diff
@@ -9,6 +9,6 @@
     </ItemGroup>
 
     <ItemGroup>
-        <PackageReference Include="TestProject" Version="#{NBGV_NuGetPackageVersion}#" />
+        <PackageReference Include="TestProject" Version="#{DependencyVersion}#" />
```
jmezach, line 12: What if we were to define two `PackageReference`'s here, one with `Condition="'$(UseFloatingVersions)'=='True'"` and one with `!=True` and then have two different tokens (one using a floating version and one without). Then we would only have to add `/p:UseFloatingVersions=True` to the `dotnet build` command depending on which SDK we're currently using?

jmezach (changes requested):
Looks good to me overall. I want to dig into the binary log to make sure it is working as expected before we merge this though.

jmezach:
Just out of curiosity, have you checked if adding `GeneratePathProperty=False` on the `PackageReference` forces the property not to be generated (even though the package has a `tools` folder)? Then we could use that to test the fallback scenario as well.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: Okay, sure. I wasn't sure if it was an example or a real-world use case, although I suppose it is quite specific 😄

jeffrosenberg:
> Just out of curiosity, have you checked if adding `GeneratePathProperty=False` on the `PackageReference` forces the property not to be generated (even though the package has a `tools` folder)?

Just tried this, and it doesn't work, but I see your point about making sure to test the fallback.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
jeffrosenberg: Are you thinking that `UseFloatingVersions` property would be just for the test project? If so, I think that could make sense. I'd probably tweak the name though, as both versions technically use floating versions, it's just a question of single vs. double wildcards.

jeffrosenberg:
Haha, I figured out that I can apparently test the fallback scenario by redefining the `$(Pkg<package-id>)` property:

`dotnet build ./test/TestProjectWithPackageReference/TestProjectWithPackageReference.csproj -c Release /p:PkgTestProject=""`

I'll work on adding this to the ci tests.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
jmezach: Yeah, sure, I think that should be something we'd use for testing only. I don't expect users to do it this way.

jeffrosenberg:
@jmezach okay, I've made the following changes to the tests, and all are passing:

* Use a conditional property to manage different test scenarios
  * I decided to call it `TestWildcard`, which can be set to "Single", "Double", or left empty
* Run tests on both floating and exact versions
  * One subset of tests for exact versions should be using the fallback method to derive the path
  * I wanted to add a negative test for this as well, i.e. a test expecting failure, but that doesn't really work with the current setup
* Add a flag to treat SQL71502 warnings as errors
  * These are the warnings that pop up when the dependencies aren't resolved correctly. 
  * On my last PR, we were scanning for this scenario manually. This flag should fail the build instead.

test/TestProjectWithSDKRef/TestProjectWithSDKRef.csproj
```diff
@@ -9,6 +9,8 @@
     </ItemGroup>
 
     <ItemGroup>
-        <PackageReference Include="TestProject" Version="#{NBGV_NuGetPackageVersion}#" />
+        <PackageReference Include="TestProject" Condition="'$(TestWildcard)'=='Double'" Version="1.*-*" />
```
jmezach, line 12: Now that I see this I'm thinking, can't we specify the Version we want to use entirely from the command line and test our various scenario's that way? So we just put `Version="$(DependencyVersion)"` in here (with a single `PackageReference` element again) and then control the `DependencyVersion` property from the command line. That would make this a lot cleaner imho, but I am sorry for changing my mind about this every time I see it ;).

jeffrosenberg:
I pushed one small change, just cleaning up the names of the test steps now that they don't have to be separated out by OS. I'm good to merge this one if you guys are. Thanks as always for working with me on this! I'm getting really close to finalizing the integration of this SDK into the database builds at my company! 🙌

jmezach:
@jeffrosenberg Thanks for your contribution. Really appreciate it. And great to hear that you're working on integrating this at your company.

Issue #41, opened 2020-08-25
[Question] How can I load the modified project back into VS2019 for editing?


shawty:
Like many people trying to build dotnet core projects on linux, that include SQL server projects, I ended up here after some internet searching and research.

Like others, I'm editing in Visual Studio, checking into GitSVN then having JenkinsCI build my project on a linux target.

Before finding this project, this was my issue

![image](https://user-images.githubusercontent.com/1876454/91228631-78360f80-e720-11ea-8c3f-30c4489b1d88.png)

After following the instructions on your main page, and replacing my project file with that shown (The sql server project was created from the VS2019 template) I now find I can build the project from the command line using "dotnet build", but I can no longer load the project into Visual Studio 2019 for editing, when I try I just get the following:

![image](https://user-images.githubusercontent.com/1876454/91228851-d7941f80-e720-11ea-80bc-5124098b04aa.png)

How can I load the project into VS for editing, but still use your project to enable me to build my solution without the presence of the project breaking the build.

ErikEJ:
Could you share your .csproj ?

shawty:
Yea sure Erik, but there's nothing special in it, it's just the standard template one. (sqlproj) I think it is.

![image](https://user-images.githubusercontent.com/1876454/91288610-19f14700-e789-11ea-87ec-9c1ab349d940.png)

![image](https://user-images.githubusercontent.com/1876454/91288686-32f9f800-e789-11ea-9518-2813dee519bf.png)

![image](https://user-images.githubusercontent.com/1876454/91288728-3f7e5080-e789-11ea-8edb-8438a44429c5.png)

![image](https://user-images.githubusercontent.com/1876454/91288845-6046a600-e789-11ea-81f0-a51f7a1b4a68.png)

This ^^^^^ above, plus the attached ".sqlproj" file, is the original project files

[st_database.sqlproj.zip](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/files/5128985/st_database.sqlproj.zip)

I then replaced the sqlproj file created by the template with 

![image](https://user-images.githubusercontent.com/1876454/91289197-e95ddd00-e789-11ea-9460-540fe00f21dc.png)

as suggested in this projects readme.

That got me to the point where I can do this

![image](https://user-images.githubusercontent.com/1876454/91289293-1611f480-e78a-11ea-9451-9011a1750037.png)

but where reloading it into VS19, gives me this

[...]

PR #42, opened 2020-08-26
Improve docs

ErikEJ:
to address #41


jmezach:
Perhaps we should add the note about having a "companion" project if you want to keep the SSDT features?

PR #52, opened 2020-09-06
Add .NET 5.0 support

jmezach:
This adds support for .NET 5 previews so that folks having the preview installed can still build their projects without needing to put a `global.json` somewhere. Essentially this means that we're now building the `DacpacTool` for .NET 5 as well and include that into the `MSBuild.Sdk.SqlProj` package. At build time we pick the appropriate version depending on the .NET runtime included with the SDK.

I've also extended the CI pipeline to start building with .NET 5.0 as well. So far this seems to work just fine. We'll need to keep this up-to-date until the stable release hits in november.

Fixes #52

### DIFF
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;netcoreapp5.0</TargetFrameworks>
     <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
--- a/src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj
+++ b/src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj
@@ -31,7 +31,7 @@
 
   <Target Name="IncludeDacpacTool" AfterTargets="Build">
     <PropertyGroup>
-      <_DacpacToolSupportedTfms>netcoreapp2.1;netcoreapp3.1</_DacpacToolSupportedTfms>
+      <_DacpacToolSupportedTfms>netcoreapp2.1;netcoreapp3.1;netcoreapp5.0</_DacpacToolSupportedTfms>
     </PropertyGroup>
     <ItemGroup>
       <DacpacToolSupportedTfms Include="$(_DacpacToolSupportedTfms)" />
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Setup .NET Core
       uses: coderpatros/setup-dotnet@sxs
       with:
-        dotnet-version: 2.1.809,3.1.401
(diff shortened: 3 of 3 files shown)


jmezach:
@ErikEJ Could you give this a quick review?

src/DacpacTool/DacpacTool.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.1;netcoreapp5.0</TargetFrameworks>
```
ErikEJ, line 5: net5.0 ??

src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj
```diff
@@ -31,7 +31,7 @@
 
   <Target Name="IncludeDacpacTool" AfterTargets="Build">
     <PropertyGroup>
-      <_DacpacToolSupportedTfms>netcoreapp2.1;netcoreapp3.1</_DacpacToolSupportedTfms>
+      <_DacpacToolSupportedTfms>netcoreapp2.1;netcoreapp3.1;netcoreapp5.0</_DacpacToolSupportedTfms>
```
ErikEJ, line 34: net5.0 ??

ErikEJ:
Had a couple of comments..

jmezach:
@ErikEJ Okay, I don't see them? Am I missing something?

ErikEJ (changes requested):

ErikEJ:
Sorry, could not figure out the UX in the GitHub app

src/DacpacTool/DacpacTool.csproj
jmezach: Yeah, that is what I had initially, but it conflicts with [this piece of code](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/192c4e992057ff1c4f4932d78dd0a50b522f95b0/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L100). I guess we could fix that by adding a `Condition` for 5.0 and higher and a separate one for lower than 5.0. More ideally though we would have the TFM of the running MSBuild and use that, but I'm not sure if that's available.

ErikEJ: Fair enough

ErikEJ (approved):

src/DacpacTool/DacpacTool.csproj
jmezach: @ErikEJ Looks like there is a `MicrosoftNETBuildTasksTFM` which is defined [here](https://github.com/dotnet/sdk/blob/8288f55cf35fb642bf2838bd11b6e17d884f019f/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Common.targets#L18). It should be available in every project, but I don't know if it is the right property to use. Looking at the history it seems to only have changed recently from `netcoreapp2.1` to `net5.0`, but it doesn't appear to have ever been `netcoreapp3.1`.

To be honest, I think it would be fine to keep it like it is currently. Unless the `netcoreapp5.0` TFM is going to be removed before .NET 5 ships. Do you happen to know if that's the case?

ErikEJ: I think it will continue to be available, at least for 5.0

PR #55, opened 2020-09-07
Improve display of validation errors

ErikEJ:
Not able to get any file or line info, though

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ In order to get started contributing code, make sure you have the following inst
 
 This project is made up of a [command line tool](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/tree/master/src/DacpacTool) that does most of the heavy lifting and an [accompanying NuGet package](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/tree/master/src/MSBuild.Sdk.SqlProj) that puts it all together. When you make changes to the `DacpacTool` you'll want to run a `dotnet build` in the `src/MSBuild.Sdk.SqlProj` folder. This ensures you're using the latest bits when building any of the [test projects](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/tree/master/test). There is a set of [unit tests](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/tree/master/test/DacpacTool.Tests) for the command line tool as well. For new features it is recommended to add new test projects in the `test` folder to ensure that the functionality works as intended.
 
-If you want to debug the project locally we recommend making a change to the Main method in [Program.cs](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/src/DacpacTool/Program.cs) by adding `System.Diagnostics.Debugger.Launch()` as the first statement. This should provide you with the opportunity to attach a debugger to the tool so that you can step through.
+If you want to debug the `DacpacTool` locally you can run `dotnet build /p:MSBuildSdkSqlProjDebug=True` on a project that uses the SDK. This will trigger the `DacpacTool` to wait for a debugger to attach before doing any actual work. Note that this only works when the `DacpacTool` is compiled for the `Debug` configuration.
 
 When you're ready to make code changes, follow these steps:
 1. Create your own fork of the code
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,9 +28,23 @@ In order to get started contributing code, make sure you have the following inst
 * .NET Core 3.1 SDK
 * Optionally: A local [SQL Server](https://www.microsoft.com/en-us/sql-server/sql-server-2019) (possibly as a [Docker container](https://hub.docker.com/_/microsoft-mssql-server))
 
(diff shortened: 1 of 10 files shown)


jmezach:
I did some digging and it seems that the `DacModelMessage` class doesn't expose those details. They are available on the `DataSchemaError` class, but that is internal as with pretty much everything from DacFx which is a bit of bummer.

ErikEJ:
That project needs to be open sourced! 

@vickyharp ?

jmezach:
@ErikEJ I had another look this mroning and I noticed that the `TSqlModel` class exposes `GetModelErrors()` method which exposes a list of `DacModelError` instances which I believe contains the information we are looking for here. Looking at the implementation I don't think we even need to call `Validate()` first, which is a bit weird, but I do think we can get richer diagnostics including source information using that method.

ErikEJ:
So we should redo validate, or just use GetModelErrors if there are any validation messages?

jmezach:
Not sure. Guess we'll have to see if `Validate()` returns the same messages as `GetModelErrors()` or not. Looking at the decompiled sources it seems that they both have their unique code paths but similar implementations.

jmezach:
I just tried giving `GetModelErrors()` a try but it seems that it returns an empty list at least for our test project which includes an error. Guess we'll have to do some more digging to figure out what will actual result in a "model error".

jmezach:
I've managed to get to the validation errors through some reflection magic and that gives us the TSQL object that has the validation error including a line number and column. It does not however give us the source file that defined that object in the first place. That makes sense since we are essentially reading the source file from disk and then passing the script as is to the `TSqlModel`. Perhaps there's a way to tell the `TSqlModel` to read the script file itself so that it knows where that object has been defined.

ErikEJ:
Maybe use this: øhttps://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.dac.model.tsqlmodel.addorupdateobjects?view=sql-dacfx-150#Microsoft_SqlServer_Dac_Model_TSqlModel_AddOrUpdateObjects_System_String_System_String_Microsoft_SqlServer_Dac_Model_TSqlObjectOptions_

jmezach:
@ErikEJ Yes, that is what I'm using now. I had to do some more reflection gymnastics to actually get to that data though but I think I've managed to make this work now. I now get full source information out of the build process. I'll have to test this on Windows to see if this works good with Visual Studio as well, but I'm confident that it'll work.

I also snug in a nice improvement to make our life easier. You can now run `dotnet build /p:MSBuildSdkSqlProjDebug=True` which will have the `DacpacTool` wait for a debugger to attach before continuing. Of course this is only included in the Debug configuration and not actually shipping in the package.

jmezach:
@ErikEJ Just tried a beta version of this in Visual Studio and it looks like warnings and errors show up nicely in Visual Studio with a mapping to the source file where the error/warning occurs:

Screenshot 2020-09-13 at 14 35 03

jmezach:
@ErikEJ @jeffrosenberg Could you both give this a review?

test/DacpacTool.Tests/ExtensionsTest.cs
```diff
@@ -31,5 +31,27 @@ public void CanAddReference()
...
+
+            // Act
+            var modelValidationErrors = model.GetModelValidationErrors(Enumerable.Empty<string>());
+
+            // Assert
+            modelValidationErrors.ShouldNotBeEmpty();
+
+            var error = modelValidationErrors.First();
+            error.Severity.ShouldBe(Microsoft.SqlServer.Dac.Model.ModelErrorSeverity.Error);
+            error.SourceName.ShouldBe("../../../../../test/TestProjectWithErrors/Procedures/csp_Test.sql");
```
ErikEJ, line 54: Maybe add an assert for ToString() ?

jmezach: Done ;).

jmezach (approved):

ErikEJ:
LGTM

CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ In order to get started contributing code, make sure you have the following inst
...
 
-If you want to debug the project locally we recommend making a change to the Main method in [Program.cs](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/src/DacpacTool/Program.cs) by adding `System.Diagnostics.Debugger.Launch()` as the first statement. This should provide you with the opportunity to attach a debugger to the tool so that you can step through.
+If you want to debug the `DacpacTool` locally you can run `dotnet build /p:MSBuildSdkSqlProjDebug=True` on a project that uses the SDK. This will trigger the `DacpacTool` to wait for a debugger to attach before doing any actual work. Note that this only works when the `DacpacTool` is compiled for the `Debug` configuration.
```
jeffrosenberg, line 33: Just one quick question on this - when you say "this only works when the `DacpacTool` is compiled for the `Debug` configuration", exactly what does that involve? Here's how I interpret it, but let me know if I'm understanding correctly:

1. The user needs to build `DacpacTool` (and potentially the whole Sdk) locally in the `Debug` configuration
2. The Sdk compiled for `Debug` needs to be installed in the local Nuget repository
3. The project to be debugged needs its Sdk reference updated for the locally-build version

Am I over-complicating things, and maybe it's simpler than that? Either way, I feel like it would help to spell out more clearly exactly what the process is for getting this to work.

jmezach: Yeah, good point. I will try to make this a little bit clearer.

jmezach:
@jeffrosenberg I've tweaked the CONTRIBUTING.md a little. Does this make it clearer?

jeffrosenberg:
@jmezach yes, looks a lot clearer to me, thanks!

jeffrosenberg (approved):

jmezach:
Let's get this merged and release it as 1.7.0 together with #58 and #59.

Issue #43, opened 2020-08-26
DacpacTool invocation fails on Windows if the maximum command length is exceeded


jeffrosenberg:
Apparently there's a limit to the length of a command sent at the Windows command line ([8191 characters](https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation), possibly up to [32767 in some situations](https://support.thoughtworks.com/hc/en-us/articles/213248526-Getting-around-maximum-command-line-length-is-32767-characters-on-Windows)). Because DacpacTool is called via the command line, and each included file is sent as a separate argument, projects with either (a) a lot of objects or (b) very long paths to each object will eventually exceed this limit and fail.

I ran into this limitation and was able to work around it by creating a symbolic link to shorten my paths, but that's a workaround at best and projects with a large number of objects won't be able to do that. 

One possible solution to the problem would be to pass the list of included objects in a temporary file, rather than via the CLI.

jeffrosenberg:
As mentioned, I've been able to work around this, so it's a low-priority issue for me personally, but I wanted to make sure it got logged so it could be dealt with at some point.

jmezach:
I guess we could write a list of files that are part of the project to a temporary file in the `obj` folder and then read that file from the `DacpacTool`. Shouldn't be too hard to do with the MSBuild built-in `WriteLinesToFile` task. If anyone's willing to sent a PR for this that would be great!

jeffrosenberg:
I'll take care of this at some point. This isn't a really high priority for me, but as you mentioned, it shouldn't be too difficult, so I'll try to get a PR in for it this week or next.

jmezach:
That would be awesome! Really appreciate all the work you've been putting into this.

jmezach:
Version 1.6.1 that contains the fix for this is now up on NuGet.org.

PR #56, opened 2020-09-08
Use an intermediate file containing each object definition

jeffrosenberg:
I've left the existing code for `--input`, just in case anyone is trying to call the DacpacTool CLI directly, but added `--inputfile` and changed `Sdk.targets` to use that instead. The new intermediate file is added to the `obj/` directory.

### DIFF
--- a/src/DacpacTool/BuildOptions.cs
+++ b/src/DacpacTool/BuildOptions.cs
@@ -9,7 +9,6 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public string Version { get; set; }
         public FileInfo Output { get; set; }
         public SqlServerVersion SqlServerVersion { get; set; }
-        public FileInfo[] Input { get; set; }
         public string InputFile { get; set; }
         public string[] Reference { get; set; }
         public string[] Property { get; set; }
--- a/src/DacpacTool/BuildOptions.cs
+++ b/src/DacpacTool/BuildOptions.cs
@@ -9,7 +9,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public string Version { get; set; }
         public FileInfo Output { get; set; }
         public SqlServerVersion SqlServerVersion { get; set; }
-        public string InputFile { get; set; }
+        public FileInfo InputFile { get; set; }
         public string[] Reference { get; set; }
         public string[] Property { get; set; }
         public string[] SqlCmdVar { get; set; }
--- a/src/DacpacTool/BuildOptions.cs
+++ b/src/DacpacTool/BuildOptions.cs
@@ -10,6 +10,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public FileInfo Output { get; set; }
         public SqlServerVersion SqlServerVersion { get; set; }
         public FileInfo[] Input { get; set; }
+        public string InputFile { get; set; }
         public string[] Reference { get; set; }
         public string[] Property { get; set; }
         public string[] SqlCmdVar { get; set; }
--- a/src/DacpacTool/Program.cs
+++ b/src/DacpacTool/Program.cs
@@ -18,7 +18,8 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 new Option<string>(new string[] { "--version", "-v" }, "Version of the package"),
                 new Option<FileInfo>(new string[] { "--output", "-o" }, "Filename of the output package"),
                 new Option<SqlServerVersion>(new string[] { "--sqlServerVersion", "-sv" }, () => SqlServerVersion.Sql150, description: "Target version of the model"),
-                new Option<FileInfo[]>(new string[] { "--input", "-i" }, "Input file name(s)"),
+                new Option<FileInfo[]>(new string[] { "--input", "-i" }, "Input file name(s)"), // DEPRECATED: Prefer InputFile instead
+                new Option<string>(new string[] { "--inputfile", "-f" }, "Text file listing all input files"),
                 new Option<string[]>(new string[] { "--reference", "-r" }, "Reference(s) to include"),
                 new Option<FileInfo>(new string[] { "--predeploy" }, "Filename of optional pre-deployment script"),
                 new Option<FileInfo>(new string[] { "--postdeploy" }, "Filename of optional post-deployment script"),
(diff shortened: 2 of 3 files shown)


src/DacpacTool/Program.cs
```diff
@@ -97,6 +99,23 @@ private static int BuildDacpac(BuildOptions options)
...
+            {
+                if (File.Exists(options.InputFile))
+                {
+                    foreach (var line in File.ReadLines(options.InputFile))
+                    {
+                        FileInfo inputFile = new FileInfo(line); // Validation occurs in AddInputFile
+                        packageBuilder.AddInputFile(inputFile);
+                    }
+                }
+                else
```
jeffrosenberg, line 113: This path shouldn't be possible, since `options.InputFile` is created by the SDK, but I wanted to make sure it was handled just in case.

src/DacpacTool/BuildOptions.cs
```diff
@@ -10,6 +10,7 @@ public class BuildOptions
         public FileInfo Output { get; set; }
         public SqlServerVersion SqlServerVersion { get; set; }
         public FileInfo[] Input { get; set; }
+        public string InputFile { get; set; }
```
ErikEJ, line 13: No one should be using this directly, so maybe just remove?

src/DacpacTool/Program.cs
```diff
@@ -18,7 +18,8 @@ static async Task<int> Main(string[] args)
                 new Option<string>(new string[] { "--version", "-v" }, "Version of the package"),
                 new Option<FileInfo>(new string[] { "--output", "-o" }, "Filename of the output package"),
                 new Option<SqlServerVersion>(new string[] { "--sqlServerVersion", "-sv" }, () => SqlServerVersion.Sql150, description: "Target version of the model"),
-                new Option<FileInfo[]>(new string[] { "--input", "-i" }, "Input file name(s)"),
+                new Option<FileInfo[]>(new string[] { "--input", "-i" }, "Input file name(s)"), // DEPRECATED: Prefer InputFile instead
```
ErikEJ, line 21: Remove?

src/DacpacTool/Program.cs
```diff
@@ -89,6 +90,7 @@ private static int BuildDacpac(BuildOptions options)
             }
 
             // Add input files
+            // DEPRECATED: Prefer InputFile instead -- Input can cause the maximum command size to be exceeded on Windows
```
ErikEJ, line 93: Remove ?

src/DacpacTool/BuildOptions.cs
jeffrosenberg: I think you mean this comment to be on the line above, `public FileInfo[] Input`, right? I'm fine removing if you guys both agree, I just personally prefer to err on the side of not removing previously-working functionality.

ErikEJ: I prefer to avoid the cost of keeping this old deprecated code alive.

jmezach: I think I'm with @ErikEJ on this. Rather not maintain code that isn't being used anymore. I'll try to give this a review this evening as well.

jeffrosenberg: Okay, no problem, I've pushed a change to remove the deprecated code.

ErikEJ:
LGTM

src/DacpacTool/Program.cs
jmezach: Yeah, I think it is a good idea in case some one does want to invoke the tool manually for some reason.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -146,13 +146,19 @@
       </PropertyNames>
       <IncludedDacpacReferenceFiles Include="@(DacpacReference->'%(DacpacFile)')" />
     </ItemGroup>
+    <!-- Write the list of input files to a file to be consumed by the dacpac tool -->
+    <Message Importance="Low" Text="Writing input files to $(IntermediateOutputPath)$(MSBuildProjectName)InputFiles.txt" />
```
jmezach, line 150: The filename in the message doesn't exactly match the actual filename.

src/DacpacTool/Program.cs
```diff
@@ -18,7 +18,7 @@ static async Task<int> Main(string[] args)
                 new Option<string>(new string[] { "--version", "-v" }, "Version of the package"),
                 new Option<FileInfo>(new string[] { "--output", "-o" }, "Filename of the output package"),
                 new Option<SqlServerVersion>(new string[] { "--sqlServerVersion", "-sv" }, () => SqlServerVersion.Sql150, description: "Target version of the model"),
-                new Option<FileInfo[]>(new string[] { "--input", "-i" }, "Input file name(s)"),
+                new Option<string>(new string[] { "--inputfile", "-i" }, "Text file listing all input files"),
```
jmezach, line 21: Is there a specific reason you've changed this from a `FileInfo` to a `string`? I believe this can still be a `FileInfo` object since it is still referring to a file on disk right?

jmezach:
Looks good to me. Just a couple of comments.

jmezach:
@jeffrosenberg I believe this fixes #43 right?

jeffrosenberg:
@jmezach correct, this fixes #43. I missed that in the original commit message when the PR was created, but it's in there now.

src/DacpacTool/Program.cs
jeffrosenberg: I feel like there was, but now it doesn't make much sense to me. 😄  I'll change this and confirm it works properly.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: Good catch, thanks, sloppy copypasta on my part.

jmezach: Haha, no problem. Happens to me all the time ;).

jmezach:
FYI We just passed 10K downloads on NuGet.org ;).

src/DacpacTool/Program.cs
jeffrosenberg: Seems to be working fine as `FileInfo`, and I agree that makes more sense. There's still a small change -- it's now `Option<FileInfo>` (singular) rather than `Option<FileInfo[]>` (array)

jmezach:
This is looking good. @ErikEJ @jeffrosenberg I would say we ship this as 1.6.1 since it is really just a small bug fix, or we can combine this with some of the other things and ship a 1.7.0. Any preference?

jeffrosenberg:
I think 1.6.1 makes sense to me, agreed that it's just a small bug. It's possible I'm the only one who's even run into this yet 😄

ErikEJ:
Makes sense as 1.6.1, just a bug fix

jmezach:
Alright. I'll go ahead and merge this to master and then cherry-pick it to `release/1.6.0` and push an updated version number. Should be up on NuGet.org soon.

PR #58, opened 2020-09-13
Add additional item templates and schema name support

jmezach:
This adds two new item templates to the template pack introduced in 1.4.0, namely user-defined data types and user-defined table types. We are using the latter quite a lot. I've also added a schemaName parameter to all of the item templates so that you can run `dotnet new sproc -n MyStoredProcedure -s MySchema` and immediately have a stored procedure with the correct name and schema created for you.


ErikEJ (approved):

jeffrosenberg (approved):
LGTM!

Issue #23, opened 2020-05-07
Support for merging SQLCMD scripts


jmezach:
In version 1.1.0 we've added support for pre- and post deployment scripts (see #9). Unfortunately we do not yet support including other scripts from those scripts using the `:r OtherScript.sql` syntax.

Some initial work has been done to support this (see `feature/sqlcmd-file-merging` branch), but it has some limitations in that it doesn't work cross-platform. This is due to the fact that the SQL libraries we're currently using seems to uppercase the path, which doesn't work on platforms that have a case-sensitive file system (ie. Linux). Ideally we would like to keep this library cross-platform, so we'll need to figure out how to work around this.

ltk-mpadelsky:
Is this all that is holding up merging in https://github.com/jmezach/MSBuild.Sdk.SqlProj/tree/feature/sqlcmd-file-merging ?
I'm wondering if you could just release this knowing that reference path resolution within DacFX is not working cross-platform yet (or I should say, converting given the platform its building  and resolving on).

I notice the DACExtensions project you commented on hasnt been worked on for about 3 years, which leaves me little hope that they'll resolve this anytime soon.

I'm wondering if a short-term solution would be to do something like this.

- Include 2 post/pre scripts, and ONLY include reference paths in them
- 1 set of pre/post scripts is named `.windows`. and the other `.linux.` with proper pathing syntax in both.
- developers could build for their single environment, or then use csproject conditionals to determine which one should be the included pre/post scripts

Does that make sense?  I haven't tried proving this out on my side yet to know if that's a viable solution or not.

I just know as it stands now, this SDK doesn't support pre/post at all (in v 1.1.0), so it would be nice to have SOMETHING sooner than later.

ErikEJ:
1.1.0 DOES support pre and post scripts, what makes you think otherwise? They must just be  single scripts, and reside in the correctly named folders

ltk-mpadelsky:
> 1.1.0 DOES support pre and post scripts, what makes you think otherwise? They must just be single scripts, and reside in the correctly named folders

Sorry, @ErikEJ !  I should've proof read my response a little better.  I edited it above.
I meant to say that it doesn't support resolving the references in those scripts!

jmezach:
[...]

PR #59, opened 2020-09-14
Add CoreCompile inputs so they'll be checked for incremental changes

### LINKED ISSUE #48: Incremental build ignores changes to pre/post deploy script
It seems that the incremental build support added in #34 ignores any changes made to pre- and/or post-deploy scripts. They are not part of the inputs, thus the `.dacpac` isn't being regenerated. This is probably also the case for the `.refactorlog` file added in #45.

jeffrosenberg:
As far as I can tell, this seems to be all we need to fix #48. [Here's the relevant documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-build-incrementally?view=vs-2019)

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -138,7 +138,7 @@
   <!--
     Performs the actual compilation of the input files (*.sql) into a .dacpac package by calling into a command line tool to do the actual work.
   -->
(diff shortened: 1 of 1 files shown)


jmezach:
Yes, this will fix #48 at least until we implement #23. When we implement that we'll need to figure out the list of included files and make sure we rebuild if any of those files have changed. That will be a little bit harder to implement, but I think this is good to have for now.

jmezach (approved):

jmezach:
@ErikEJ Could you give this a quick review as well?

ErikEJ (approved):

Issue #39, opened 2020-08-21
Pass SQLCMD variables at build time


lewism36:
Is it possible to reference objects defined in a referenced package?  In a view for example.

I can't see how to specify the DatabaseSqlCmdVariable or an equivalent, like when adding a project reference to a sqlproj, but I'm also not sure if the SqlCmd variable is being expanded with the default value anyway.

Without this the build fails with validation errors e.g. 

`1>Target CoreCompile:`
`1>  Using package name Database4 and version 1.0.0`
`1>  Using SQL Server version Sql150`
`1>  Adding reference to {}\.nuget\packages\database3\1.0.0\tools\Database3.dacpac`
`1>  Adding SqlCmd variable Database3`
`1>  Adding {}\Database4\View2.sql to the model`
`1>  EXEC : error SQL71561: Error validating element [dbo].[View2]: SqlView: [dbo].[View2] has an unresolved reference to object [$(Database3)].[dbo].[Table2].[Id].`
`1>  EXEC : error SQL71561: Error validating element [dbo].[View2]: SqlView: [dbo].[View2] has an unresolved reference to object [$(Database3)].[dbo].[Table2].`
`1>  EXEC : error SQL71561: Error validating element [dbo].[View2].[Id]: SqlComputedColumn: [dbo].[View2].[Id] has an unresolved reference to object [$(Database3)].[dbo].[Table2].[Id].`
`1>  Found 3 error(s), skip building package`

lewism36:
@jmezach Thanks for looking into this.  I've shared an example in [here](https://github.com/lewism36/database-projects).

jmezach:
Looking at the example I think there's actually two things that are relevant here. The first is that this requires the value of SQLCMD variables to be passed at build time. For example, given the following excerpt from the project file:

```xml
<ItemGroup>
  <SqlCmdVariable Include="Database1">
    <DefaultValue>Database1</DefaultValue>
    <Value>$(SqlCmdVar__1)</Value>
  </SqlCmdVariable>
</ItemGroup>
```

[...]

PR #61, opened 2020-09-16
Handle includes in pre- and post-deployment scripts (fixes #23)

### LINKED ISSUE #23: Support for merging SQLCMD scripts

jeffrosenberg:
I'm expecting there may be a lot of feedback and proposed changes here, as this goes beyond my comfort zone for dotnet development. I've broken this into a few commits in case we want to replace some of the work.


src/DacpacTool/PackageBuilder.cs
```diff
@@ -305,7 +306,8 @@ private void WritePart(FileInfo file, Package package, string path)
 
             using (var stream = part.GetStream())
             {
-                var buffer = Encoding.UTF8.GetBytes(File.ReadAllText(file.FullName, Encoding.UTF8) + Environment.NewLine + "GO" + Environment.NewLine);
+                var parser = new ScriptParser(file.FullName, new IncludeVariableResolver());
+                var buffer = Encoding.UTF8.GetBytes(parser.GenerateScript());
```
jeffrosenberg, line 310: Use the `ScriptParser` to generate the full script after merging any includes.

src/DacpacTool/PackageInspector.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.IO;
+using System.Collections.Generic;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool
+{
+    public sealed class PackageInspector
```
jeffrosenberg, line 7: I decided to make a parallel class to `PackageBuilder` and `PackageDeployer`, but I'm not married to this. The code here could easily be added to `PackageBuilder`, but I thought it would be cleaner to have a class for each possible invocation of `DacpacTool` from MSBuild.

src/DacpacTool/Program.cs
```diff
@@ -33,6 +33,16 @@ static async Task<int> Main(string[] args)
             };
             buildCommand.Handler = CommandHandler.Create<BuildOptions>(BuildDacpac);
 
+            var inspectCommand = new Command("inspect")
```
jeffrosenberg, line 36: New invocation of `DacpacTool` to get a list of includes for checking incremental builds. I'm not sure if anyone has thoughts on what the command should be called, I don't love "inspect"... maybe "inspect-includes" or "get-includes"?

jmezach, line 36: This is starting to become a little bit messy now. We might want to move to a model where each command is implement in its own class. I'll try to come up with a design for that, let's keep it like it is for now though.

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,95 @@
...
+            _scriptBuilder.Append(batchText);
+            _scriptBuilder.Append("\nGO\n");
+            return BatchParserAction.Continue;
+        }
+
+        BatchParserAction ICommandHandler.Include(TextBlock includeBlock, out TextReader stream, out string newFilename)
+        {
+            includeBlock.GetText(true, out string includedFileName, out LineInfo lineInfo);
+
+            // HACK: Work around access limitations to get access to the underlying tokens
```
jeffrosenberg, line 67: I hate this section and I'm open to better ways of doing this. The problem is that just storing the last value of `_currentFilePath` doesn't work for certain recursive arrangements of includes -- see `test/TestProjectWithPrePost/Pre-Deployment/Script.PreDeployment.NestedRelativePaths.sql` for an example.

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,95 @@
...
+            else
+            {
+                sourceFile = Path.Combine(Environment.CurrentDirectory, sourceFile);
+                _currentFilePath = Path.GetDirectoryName(sourceFile);
+            }
+
+            _parser = new Parser(this, variableResolver, new StreamReader(sourceFile), sourceFile);
+        }
+
+        protected void Parse()
```
jmezach, line 34: Shouldn't this be private?

jeffrosenberg:
I've added a number of unit tests, but I think I'll look at adding a postdeployment script to one of the pipeline tests as well. I'll work on that while waiting for review.

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,95 @@
...
+            // HACK: Work around access limitations to get access to the underlying tokens
+            var tokens = 
+                includeBlock.GetType()
+                            .GetField("tokens", BindingFlags.NonPublic | BindingFlags.Instance)
+                            .GetValue(includeBlock) as IEnumerable<Token>;
+            var token = tokens.FirstOrDefault();
+            if (token != null)
+            {
+                _currentFilePath = Path.GetDirectoryName(token.Filename);
+            }
```
jmezach, line 76: Why do we need access to the tokens? Isn't everything we need here contained in `includedFileName`?

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,95 @@
...
+            _includedFileNames.Add(includedFileName);
+
+            stream = new StreamReader(includedFileName);
+            newFilename = includedFileName;
+            return BatchParserAction.Continue;
+        }
+
+        BatchParserAction ICommandHandler.OnError(Token token, OnErrorAction action)
+        {
+            return BatchParserAction.Continue;
```
jmezach, line 92: Do we want to write something to the console here perhaps? Can we get any details on the error?

jmezach (changes requested):
@jeffrosenberg This is off to a great start. I do have a couple of comments but it is looking pretty good already.

src/DacpacTool/Program.cs
jmezach: Yeah, I was thinking of having a pre-process command that wouldn't just list the included files but would also generate the entire script to the `obj` folder so that we can then use that to create the final package. But then you wouldn't have the incremental build support. But maybe `collect-includes` would be an appropriate name?

src/DacpacTool/PackageInspector.cs
jmezach: Yeah, I was thinking we might not need it all but I'm guessing it makes things a little bit easier to test. I suggested `ScriptInspector` as an alternate name in my review since it doesn't have anything to do with packages.

src/DacpacTool/PackageBuilder.cs
jmezach: That makes total sense to me :).

src/DacpacTool/DacpacTool.csproj
```diff
@@ -7,9 +7,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SqlServer.DACFX" Version="150.4826.1" />
+    <PackageReference Include="Microsoft.SqlServer.DACFX" Version="150.4861.1-preview" />
```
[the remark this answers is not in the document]
jeffrosenberg: Yeah, I agree. This bit was from your initial commit for #23 -- I tested reverting it, because I didn't like depending on the preview version, but it caused the build to fail 😕

src/DacpacTool/ScriptParser.cs
jeffrosenberg: The tokens are the only place where we can get access to the path information tracked by `Microsoft.SqlTools.ServiceLayer.BatchParser.Lexer.inputStack`. The problem is that, within recursively-included files, each included filename is relative to the current path of the file it's included within.

This was the only way I could resolve the problem posed by setups like `test/TestProjectWithPrePost/Pre-Deployment/Script.PreDeployment.NestedRelativePaths.sql` -- which is based on a real-life issue I had when trying to build one of my own projects.

src/DacpacTool/ScriptParser.cs
jeffrosenberg: I think I wrote this in a moment of irritation with being unable to access any of the members of `Microsoft.SqlTools.ServiceLayer.BatchParser.Parser`. 😄  I'll set it to private.

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Yes, looks like we can, I'll add that. I'm also thinking that `Continue` wouldn't usually be the right response here, I'm going to include that in the change.

src/DacpacTool/ScriptParser.cs
jmezach: I did some digging and it turns out that through the `LineInfo` object returned by `GetText()` on the `includeBlock` has a method that we can use to get the original position from which the include was being made including the filename. I'm a little bit unsure about the argument you need to pass into that function though. I'm currently just passing in `0` which seems to work, or at least it passes the tests.

jeffrosenberg: Thanks, wow, it never occurred to me at all that `Filename` would be a member of something called `PositionStruct`, thanks for finding that!

jmezach: Yeah, it doesn't make much sense to me either. It seems that the `GetStreamPositionForOffset()` method I'm using is also called internally with a hardcoded argument of `0` so I think we can accept this as a viable option.

src/DacpacTool/IncludeVariableResolver.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.SqlTools.ServiceLayer.BatchParser;
+using Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool
+{
+    // This was just copied from Microsoft.SqlTools.ManagedBatchParser.UnitTests.BatchParser.TestVariableResolver
+    // No special functionality needed for our project
+    internal sealed class IncludeVariableResolver : IVariableResolver
```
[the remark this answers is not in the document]
jeffrosenberg: That might be the case, I'd be happy to look into #39 and explore this further. I think for the purposes of this PR, we can leave it here though?

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Agreed, I'm just going to test it quickly with my real-life project to confirm

src/DacpacTool/IncludeVariableResolver.cs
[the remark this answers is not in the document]
jmezach: Sure. We can revisit this when we tackle #39. I think we have some more work to do there on the MSBuild side of things as well because we'll need to do some conditional resolving and stuff. But we'll cross that bridge when we get there ;).

src/DacpacTool/ScriptParser.cs
jmezach: Of course! Let me know how that goes ;).

src/DacpacTool/DacpacTool.csproj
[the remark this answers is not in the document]
jmezach: We might want to pin the submodule version of `SqlToolsService` to a version that builds against a stable version of `Microsoft.SqlServer.DACFX`. When I initially looked at this I was unable to find a commit that referenced a stable version though.

src/DacpacTool/PackageInspector.cs
jeffrosenberg: Renamed to `ScriptInspector`.

src/DacpacTool/Program.cs
jeffrosenberg: Renamed to `collect-includes`.

src/DacpacTool/InspectOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.IO;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool
+{
+    public class InspectOptions
```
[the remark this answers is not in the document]
jeffrosenberg: Added a `BaseOptions` abstract class.

src/DacpacTool/Program.cs
jeffrosenberg: Agreed, but not making changes as part of this PR, per your comment.

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Set to private.

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Confirmed that this works for me, and passes all tests. Looks so much better!

src/DacpacTool/ScriptParser.cs
jeffrosenberg: @jmezach please let me know what you think of the changes to this method

src/DacpacTool/Program.cs
```diff
@@ -132,6 +142,33 @@ private static int BuildDacpac(BuildOptions options)
             return 0;
         }
 
+        private static int InspectIncludes(InspectOptions options)
+        {
+            // Wait for a debugger to attach
+            WaitForDebuggerToAttach(options.Debug);
```
jmezach, line 148: I guess we can refactor this method to accept a `BaseOptions` instance now and just pass in the options here.

src/DacpacTool/Program.cs
```diff
@@ -132,6 +142,33 @@ private static int BuildDacpac(BuildOptions options)
...
+
+            var packageInspector = new ScriptInspector();
```
jmezach, line 150: Might want to rename the variable here as well.

jeffrosenberg: oops 🤦

src/DacpacTool/ScriptInspector.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.IO;
+using System.Collections.Generic;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool
+{
+    public sealed class ScriptInspector
+    {
+        private readonly List<string> _includedFiles = new List<string>();
+        public List<string> IncludedFiles
```
jmezach, line 10: We might want to change the type of this property to `IEnumerable<string>` since it is supposed to be readonly.

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.SqlTools.Extensibility;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool
+{
+    public class ScriptParser : ICommandHandler
+    {
+        private readonly Parser _parser;
+        private bool _parsed;
+        private readonly List<string> _includedFileNames = new List<string>();
+        private StringBuilder _scriptBuilder = new StringBuilder();
```
jmezach, line 17: The StringBuilder can be made `readonly` as well.

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,86 @@
...
+        private void Parse()
+        {
+            if (! _parsed)
+            {
+                _parser.Parse();
+                _parsed = true;
+            }
+        }
+
+        public List<string> CollectFileNames()
```
jmezach, line 33: We might want to change this to `IEnumerable<string>` as well.

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            Parse();
+            return _scriptBuilder.ToString();
+        }
+
+        BatchParserAction ICommandHandler.Go(TextBlock batch, int repeatCount, SqlCmdCommand tokenType)
+        {
+            batch.GetText(true, out string batchText, out LineInfo lineInfo);
+            _scriptBuilder.Append(batchText);
+            _scriptBuilder.Append("\nGO\n");
```
jmezach, line 49: This is a bit of a magic string of course. Doesn't the included library have a constant for this?

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,86 @@
...
+        BatchParserAction ICommandHandler.OnError(Token token, OnErrorAction action)
+        {
+            // Write error to console
+            Console.Error.WriteLine($"Error encountered in {token.Filename}, line {token.Begin.Line}");
+            if (token.TokenType == LexerTokenType.Text || token.TokenType == LexerTokenType.Include)
+            {
+                Console.Error.WriteLine(token.Text);
+            }
+
+            if (action == OnErrorAction.Ignore)
```
jmezach, line 79: Perhaps this could be simplified as `return action == OnErrorAction.Ignore ? BatchParserAction.Continue : BatchParserAction.Abort;`

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            stream = new StreamReader(includedFileName);
+            newFilename = includedFileName;
+            return BatchParserAction.Continue;
+        }
+
+        BatchParserAction ICommandHandler.OnError(Token token, OnErrorAction action)
+        {
+            // Write error to console
+            Console.Error.WriteLine($"Error encountered in {token.Filename}, line {token.Begin.Line}");
```
jmezach, line 73: Could we format the output similar to how the recently added `ModelValidationError` class does it? That would at least give users traceability to where the error is occurring. We might have to come up with our own error code though. I'm wondering how SSDT handles these errors. Could you try to trigger an error with that and see how that is handled?

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -135,10 +136,35 @@
...
+      <PreDeploymentScriptArgument>@(PreDeploy->'--predeploy %(Identity)', ' ')</PreDeploymentScriptArgument>
+      <PostDeploymentScriptArgument>@(PostDeploy->'--postdeploy %(Identity)', ' ')</PostDeploymentScriptArgument>
+      <DebugArgument Condition="'$(MSBuildSdkSqlProjDebug)' == 'True'">--debug</DebugArgument>
+      <DacpacToolCommand>dotnet &quot;$(DacpacToolExe)&quot; collect-includes $(PreDeploymentScriptArgument) $(PostDeploymentScriptArgument) $(DebugArgument)</DacpacToolCommand>
+    </PropertyGroup>
+    <!-- Run it, except during design-time builds -->
+    <Message Importance="low" Text="Running command: $(DacpacToolCommand)" />
+    <Exec Command="$(DacpacToolCommand)" ConsoleToMsBuild="true" EchoOff="true" StandardOutputImportance="Low"
+          Condition="'$(DesignTimeBuild)' != 'true' AND '$(BuildingProject)' == 'true'">
+      <Output TaskParameter="ConsoleOutput" PropertyName="IncludedFiles" />
```
jmezach, line 154: Isn't there an `ItemName` attribute here as well that we can use, rather than having to convert it into an ItemGroup manually later on? Looking at the [docs for the Exec task](https://docs.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2019#parameters) it seems to be an `ITaskItem[]` property.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -41,6 +41,7 @@
       PreBuildEvent;
       ResolveReferences;
       GetTargetPath;
+      GetIncludedFiles;
```
jmezach, line 44: Is this necessary? We're already listing this target as a dependency for `CoreCompile`?

test/DacpacTool.Tests/ScriptParserTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using NSubstitute;
+using Shouldly;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool.Tests
+{
+    [TestClass]
+    public class ScriptParserTests
+    {
+        private const string testProjectPath = "../../../../TestProjectWithPrePost";
```
jmezach, line 11: Perhaps we should rename this to TEST_PROJECT_PATH to match with the general C# guidelines.

test/TestProjectWithPrePost/TestProjectWithPrePost.nuspec
```diff
@@ -0,0 +1,20 @@
...
+    <description>$description$</description>
+    <authors>$authors$</authors>
+    <owners>$authors$</owners>
+    <copyright>$copyright$</copyright>
+    <projectUrl>$projecturl$</projectUrl>
+    <tags>$tags$</tags>
+    <packageTypes>
+      <packageType name="$packagetype$" />
+    </packageTypes>
+    <repository type="git" url="https://github.com/jmezach/MSBuild.Sdk.SqlProj/" />
```
jmezach, line 15: Could you replace `jmezach` with `rr-wfm` in this URL?

src/DacpacTool/ScriptParser.cs
jeffrosenberg: There's one in another assembly, but it seems like a pretty small thing to add another assembly reference for. I could just create my own constant for it if you'd like, but the string "GO" was what we had here before.

jmezach:
@jeffrosenberg On reviewing it another time I found a couple of more things. Sorry about that ;).

src/DacpacTool/ScriptParser.cs
jmezach: Yeah, adding an assembly reference for just a constant doesn't make much sense. It would just be copied over to our assembly anyway. Perhaps we should use `AppendLine()` here though, rather than passing in the newlines explicitly.

jeffrosenberg:
@jmezach no problem! I'm about to run into a bunch of meetings, so I'll probably get the remaining changes pushed out tomorrow morning.

src/DacpacTool/ScriptParser.cs
jeffrosenberg: I've created a constant called `BATCH_SEPARATOR` in this file

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: No idea, to be honest, I'll remove it and if it still works I'll leave it removed 😄

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Agreed about `AppendLine()`, I can't remember what I was thinking when I wrote it this way but it doesn't make much sense to me looking at it now 😄

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: Confirmed I can just use `$(IncludedFiles)` instead of converting to an ItemGroup, I didn't realize that Properties could be iterated over like ItemGroups, but it seems to be working to me.

jmezach:
@ErikEJ Can you take a look at this as well?

ErikEJ:
I will this evening

src/DacpacTool/DacpacTool.csproj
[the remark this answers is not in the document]
ErikEJ: Not very fond of this... What build errors did you get?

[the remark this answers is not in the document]
jeffrosenberg: I'm not either, and I'd be happy if there was a solution that allowed us to avoid this. Here's the build output:

```
~/MSBuild.Sdk.SqlProj/src/DacpacTool/DacpacTool.csproj : error NU1605: Detected package downgrade: Microsoft.SqlServer.DACFx from 150.4861.1-preview to 150.4826.1. Reference the package directly from the project to select a different version. 
~/MSBuild.Sdk.SqlProj/src/DacpacTool/DacpacTool.csproj : error NU1605:  DacpacTool -> Microsoft.SqlTools.ManagedBatchParser -> Microsoft.SqlServer.DACFx (>= 150.4861.1-preview) 
~/MSBuild.Sdk.SqlProj/src/DacpacTool/DacpacTool.csproj : error NU1605:  DacpacTool -> Microsoft.SqlServer.DacFx (>= 150.4826.1)
  Failed to restore ~/MSBuild.Sdk.SqlProj/src/DacpacTool/DacpacTool.csproj (in 981 ms).
```

src/DacpacTool/DacpacTool.csproj
```diff
@@ -7,9 +7,13 @@
...
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20427.1" />
     <PackageReference Include="System.Security.Permissions" Version="4.7.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="../../sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/Microsoft.SqlTools.ManagedBatchParser.csproj" />
+  </ItemGroup>
+
```
ErikEJ, line 18: Why not just include the files from this folder: https://github.com/microsoft/sqltoolsservice/tree/main/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser

src/DacpacTool/Program.cs
```diff
@@ -48,7 +58,7 @@ static async Task<int> Main(string[] args)
             };
             deployCommand.Handler = CommandHandler.Create<DeployOptions>(DeployDacpac);
 
-            var rootCommand = new RootCommand { buildCommand, deployCommand };
+            var rootCommand = new RootCommand { buildCommand, inspectCommand, deployCommand };
```
ErikEJ, line 61: What is "inspect command"??

jeffrosenberg: New CLI command to collect the list of included files before running the build. Although we renamed the command "collect-includes", so I should rename this variable as well. I'll include that in my next update.

src/DacpacTool/DacpacTool.csproj
jeffrosenberg: Doesn't this need to be a reference to a project, rather than just a list of files?

jmezach: Hmm, yeah, we might be able to get away with `<Compile Include="../../sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/**/*.cs" />`. That would essentially recompile those source files into the `DacpacTool` but I would be fine with that.

jmezach: We might need to add additional `PackageReferences` to make that work though

jeffrosenberg: I'll play around with this and see if I can make it work.

jeffrosenberg: @ErikEJ @jmezach I've pushed a change that gets really close here -- everything compiles, but 1 test is failing and it's beyond my ability to understand why. It seems to be an issue with the localization resource, but I have the resource included just as `ManagedBatchParser.csproj` does.

Here's the test failure, any thoughts?

```
Test method MSBuild.Sdk.SqlProj.DacpacTool.Tests.PackageBuilderTest.AddPreDeployment_FilesExist threw exception: 
System.Resources.MissingManifestResourceException: Could not find the resource "Microsoft.SqlTools.ManagedBatchParser.Localization.SR.resources" among the resources "Microsoft.SqlTools.ManagedBatchParser.SR.resources" embedded in the assembly "DacpacTool", nor among the resources in any satellite assemblies for the specified culture. Perhaps the resources were embedded with an incorrect name.
  Stack Trace:
      at System.Resources.ManifestBasedResourceGroveler.HandleResourceStreamMissing(String fileName)
   at System.Resources.ManifestBasedResourceGroveler.GrovelForResourceSet(CultureInfo culture, Dictionary`2 localResourceSets, Boolean tryParents, Boolean createIfNotExists)
   at System.Resources.ResourceManager.InternalGetResourceSet(CultureInfo culture, Boolean createIfNotExists, Boolean tryParents)
   at System.Resources.ResourceManager.GetString(String name, CultureInfo culture)
   at Microsoft.SqlTools.ManagedBatchParser.SR.Keys.GetString(String key) in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/Localization/sr.cs:line 409
   at Microsoft.SqlTools.ManagedBatchParser.SR.get_EE_ExecutionError_VariableNotFound() in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/Localization/sr.cs:line 156
   at Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode.BatchParserSqlCmd.GetVariable(PositionStruct pos, String name) in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/ExecutionEngineCode/BatchParserSqlCmd.cs:line 72
   at MSBuild.Sdk.SqlProj.DacpacTool.IncludeVariableResolver.GetVariable(PositionStruct pos, String name) in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/IncludeVariableResolver.cs:line 24
   at Microsoft.SqlTools.ServiceLayer.BatchParser.Parser.ResolveVariables(Token inputToken, Int32 offset, List`1 variableRefs) in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs:line 661
   at Microsoft.SqlTools.ServiceLayer.BatchParser.TextBlock.GetText(Boolean resolveVariables, String& text, LineInfo& lineInfo) in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/TextBlock.cs:line 54
   at MSBuild.Sdk.SqlProj.DacpacTool.ScriptParser.Microsoft.SqlTools.ServiceLayer.BatchParser.ICommandHandler.Go(TextBlock batch, Int32 repeatCount, SqlCmdCommand tokenType) in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/ScriptParser.cs:line 47
   at Microsoft.SqlTools.ServiceLayer.BatchParser.Parser.ExecuteBatch(Int32 repeatCount) in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs:line 204
   at Microsoft.SqlTools.ServiceLayer.BatchParser.Parser.Parse() in ~/MSBuild.Sdk.SqlProj/sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser/BatchParser/Parser.cs:line 233
   at MSBuild.Sdk.SqlProj.DacpacTool.ScriptParser.Parse() in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/ScriptParser.cs:line 28
   at MSBuild.Sdk.SqlProj.DacpacTool.ScriptParser.GenerateScript() in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/ScriptParser.cs:line 41
   at MSBuild.Sdk.SqlProj.DacpacTool.PackageBuilder.WritePart(FileInfo file, Package package, String path) in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/PackageBuilder.cs:line 310
   at MSBuild.Sdk.SqlProj.DacpacTool.PackageBuilder.AddScript(FileInfo script, FileInfo outputFile, String path) in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/PackageBuilder.cs:line 299
   at MSBuild.Sdk.SqlProj.DacpacTool.PackageBuilder.AddPostDeploymentScript(FileInfo script, FileInfo outputFile) in ~/MSBuild.Sdk.SqlProj/src/DacpacTool/PackageBuilder.cs:line 89
   at MSBuild.Sdk.SqlProj.DacpacTool.Tests.PackageBuilderTest.AddPreDeployment_FilesExist() in ~/MSBuild.Sdk.SqlProj/test/DacpacTool.Tests/PackageBuilderTests.cs:line 170
```

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: Removing seems to be fine, change pushed.

test/TestProjectWithPrePost/TestProjectWithPrePost.nuspec
jeffrosenberg: Done

test/DacpacTool.Tests/ScriptParserTests.cs
jeffrosenberg: Done

src/DacpacTool/ScriptParser.cs
jeffrosenberg: @jmezach @ErikEJ I've pushed a change for this, but I'm not sure this is how we want to handle it... for now, I've basically copied the `StringBuilder` code from `ModelValidationError` into this method. I'd rather reuse `ModelValidationError`, but I don't think I can instantiate a `DacModelError`, since it doesn't have a documented constructor.

Another option would be to convert `ModelValidationError` to have a bunch of private fields that get copied in from the `DacModelError` on instantiation, then provide an overload that allows all of those fields to be passed as individual arguments, but thought I'd run it by you guys before making a change like that.

src/DacpacTool/Program.cs
```diff
@@ -132,6 +142,33 @@ private static int BuildDacpac(BuildOptions options)
             return 0;
         }
 
+        private static int InspectIncludes(InspectOptions options)
+        {
+            // Wait for a debugger to attach
+            WaitForDebuggerToAttach(options.Debug);
+
+            var packageInspector = new PackageInspector();
```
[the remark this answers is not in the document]
jeffrosenberg: Renamed to ScriptInspector. I'm keeping this class per your comment elsewhere in the review that it simplifies testing.

src/DacpacTool/DacpacTool.csproj
jmezach: @jeffrosenberg I think that has to do with the <RootNamespace> property set in the project file. I believe that is used when loading the resources. You could try setting it to `Microsoft.SqlTools.ManagedBatchParser` and see if that helps but I'm not sure if I'm happy with that solution to be honest ;).

src/DacpacTool/Program.cs
jeffrosenberg: Fixed.

src/DacpacTool/ScriptParser.cs
jmezach: I think that makes sense. When I added the `ModelValidationError` class I just designed it to fix the problem at hand, but if we're going to use it in other places I don't mind needing to do some copying of that data from the `DacModelError`.

src/DacpacTool/DacpacTool.csproj
jeffrosenberg: @jmezach sorry, I'm not clear which property you're referring to

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Great, I've pushed that change

src/DacpacTool/DacpacTool.csproj
jmezach: Ah, sorry, I meant the `RootNamespace` property. That is not set in the project file currently, but you can add it to see if that helps.

jeffrosenberg: I think I found another property to fix this -- setting the `LogicalName` for the resource fixed it for me locally, CI tests running now.

jmezach: Looks like that fixes it. Much better solution :).

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Done

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Done

src/DacpacTool/ScriptInspector.cs
jeffrosenberg: Done

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Done

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Done

src/DacpacTool/DacpacTool.csproj
```diff
@@ -6,10 +6,29 @@
     <LangVersion>8.0</LangVersion>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <SqlToolsPath>../../sqltoolsservice/src</SqlToolsPath>
+    <ManagedBatchParserPath>../../sqltoolsservice/src/Microsoft.SqlTools.ManagedBatchParser</ManagedBatchParserPath>
```
jmezach, line 11: We might want to re-use $(SqlToolsPath) here so that if we change that it will also changed the ManagedBatchParserPath.

src/DacpacTool/DacpacTool.csproj
```diff
@@ -6,10 +6,29 @@
...
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20427.1" />
     <PackageReference Include="System.Security.Permissions" Version="4.7.0" />
   </ItemGroup>
 
+  <!-- References to Microsoft.SqlTools.ManagedBatchParser -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="161.41011.9" />
+    <ProjectReference Include="$(SqlToolsPath)/Microsoft.SqlTools.Hosting/Microsoft.SqlTools.Hosting.csproj" />
+    <Compile Include="$(ManagedBatchParserPath)/Localization/*.cs" />
+      <EmbeddedResource Include="$(ManagedBatchParserPath)/Localization/sr.resx" LogicalName="Microsoft.SqlTools.ManagedBatchParser.Localization.SR.resources" />
```
jmezach, line 25: Indentation is a little bit off here.

src/DacpacTool/ModelValidationError.cs
```diff
@@ -9,34 +9,62 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
...
+        private readonly ModelErrorSeverity _severity;
+        private readonly string _prefix;
+        private readonly int _errorCode;
+        private readonly string _message;
 
         public ModelValidationError(DacModelError modelError, string sourceName)
         {
-            _modelError = modelError ?? throw new ArgumentNullException(nameof(modelError));
-            _sourceName = sourceName.Replace("MSSQL::", string.Empty);
+            if (modelError == null) throw new ArgumentNullException(nameof(modelError));
```
jmezach, line 23: Perhaps this could be simplified to `modelError = modelError ?? throw new ArgumentNullException(nameof(modelError));`

src/DacpacTool/ScriptParser.cs
```diff
@@ -0,0 +1,87 @@
...
+            stream = new StreamReader(includedFileName);
+            newFilename = includedFileName;
+            return BatchParserAction.Continue;
+        }
+
+        BatchParserAction ICommandHandler.OnError(Token token, OnErrorAction action)
+        {
+            // Write error to console
+            var error = new ModelValidationError(
+                sourceName: Path.GetFileNameWithoutExtension(token.Filename),
```
jmezach, line 73: Why `GetFilenameWithoutExtension()`?

jmezach (approved):
I had a couple more comments but with that I think we've got this pretty much done. Thanks a bunch @jeffrosenberg for your work on this. Really appreciate it.

ErikEJ (approved):

src/DacpacTool/ScriptParser.cs
jeffrosenberg: I thought about this one for a while, and I could really go either way on it. `sourceName` is meant to be the name of a TSql object, and would typically not include the extension.

src/DacpacTool/ModelValidationError.cs
jeffrosenberg: I don't follow why I would need to set `modelError = modelError`. Perhaps you mean `_modelError = modelError`, but I removed that internal `_modelError` field when I refactored this class.

src/DacpacTool/ScriptParser.cs
jmezach: I went with the full filename in #55 because that gives us traceability to Visual Studio. That way you can double click the error in the Error List and be taken to the file that is causing the error.

src/DacpacTool/ModelValidationError.cs
jmezach: It is called a [throw expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/throw#the-throw-expression) but can only happen as part of an assignment so that's why you'd do the assignment. I know it is a bit weird.

I can live with the if as well, but then I'd rather have braces around the throw rather than the current one-liner.

src/DacpacTool/ScriptParser.cs
jeffrosenberg: Okay, if the full filename gives us better functionality, I'll go with that.

src/DacpacTool/ModelValidationError.cs
jeffrosenberg: okay, got it, I'll make that change.

src/DacpacTool/DacpacTool.csproj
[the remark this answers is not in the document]
jmezach: I believe we fixed this ;).

src/DacpacTool/ScriptParser.cs
jmezach: Looking good now.

src/DacpacTool/Program.cs
jmezach: This got implemented as well.

src/DacpacTool/ScriptParser.cs
jmezach: This got added too

src/DacpacTool/Program.cs
jmezach: @ErikEJ Do you have concerns about this, or was this just a question?

jmezach:
@jeffrosenberg Did you manage to add a test project to the CI pipeline so that we can make sure that works as expected on every change?

src/DacpacTool/Program.cs
ErikEJ: Still a confusing name... Prefer collectIncludes or consolidateIncludes

jeffrosenberg: I believe it's now `collectIncludesCommand`

jmezach: It got renamed to collect-includes in a later commit ;).

jeffrosenberg:
@jmezach I didn't add a new test project, but I did add two changes to TestProjectWithSDKRef that I think should be satisfactory for CI testing, let me know if you disagree.

https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/61/files#diff-d8d31a2c069adda9b8a4b6d9ca05c661
https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/61/files#diff-65850b3cc354d04c0e4395d2478e7161

jmezach:
@jeffrosenberg Ah, yes, that will work. I was looking for a change to the CI pipeline, but that isn't necessary if we added it to the existing project.

I guess we're good to go here then. Right @jeffrosenberg @ErikEJ?

jeffrosenberg:
I think so!

ErikEJ:
Agree :shipit:

jmezach:
Awesome. I'll get this merged and released as 1.8.0. Thanks again @jeffrosenberg for your work on this. This is probably our most requested feature currently ;).

PR #72, opened 2020-10-03
Add support for project references

### LINKED ISSUE #40: Add support for ProjectReference
While MSBuild.Sdk.SqlProj supports `PackageReference` for referencing NuGet packages and automatically include their embedded `.dacpac`'s as part of the build we do not currently support `ProjectReference`. In a `.sqlproj` this is typically defined as:

```xml
<ProjectReference Include="..\Database1\Database1.sqlproj">
  <Name>Database1</Name>
  <Project>{b4a1acc8-f726-499e-a0d5-955e146dc10a}</Project>
  <Private>True</Private>
  <SuppressMissingDependenciesErrors>False</SuppressMissingDependenciesErrors>
  <DatabaseSqlCmdVariable>Database1</DatabaseSqlCmdVariable>
</ProjectReference>
```

In the case of SDK-style projects we can probably get rid of the `Project` metadata since those GUID's aren't used anymore. I'm not entirely sure what the `Name` and `DatabaseSqlCmdVariable` metadata is used for though. The `Private` metadata probably controls whether the `.dacpac` of the project being referenced is being copied to the output folder of the project that has the reference, which shouldn't be too hard to implement. The `SuppressMissingDependenciesErrors` might be a little bit more tricky and will require some research.

We'll also need to resolve the path of the `.dacpac` for the referenced project. I'm hoping that we can already get that from MSBuild somehow since it supports project to project references already.

jmezach:
In addition to package references you might also want to reference another project (possibly within the same solution). This adds support for that. Luckily most of the heavy lifting is already done for us, so all we really needed to do is to make sure we support getting a target path of the `.dacpac` without building the project and then resolving that target path into a reference parameter for the `DacpacTool`.

Fixes #40

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -80,7 +80,7 @@
   <!--
     The GetTargetPathWithTargetPlatformMoniker target uses a BeforeTargets so the only way to disable it is to override it with an empty target.
   -->
-  <Target Name="GetTargetPathWithTargetPlatformMoniker" />
+  <!--<Target Name="GetTargetPathWithTargetPlatformMoniker" />-->
 
   <Import Project="$(CustomAfterNoTargets)" Condition="'$(CustomAfterNoTargets)' != '' and Exists('$(CustomAfterNoTargets)')" />
 
@@ -109,7 +109,7 @@
     Resolves package references to .dacpac packages by enumerating all package references, resolving their associated Pkg<package-id> property to get the physical
     location of the package and then checking if a that package contains a .dacpac file inside of the tools folder.
   -->
-  <Target Name="ResolveDatabasePackageReferences">
+  <Target Name="ResolveDatabaseReferences">
     <ItemGroup> 
       <!-- Resolve all package references to their physical location first -->
       <_ResolvedPackageReference Include="%(PackageReference.Identity)">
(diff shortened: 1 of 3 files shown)


src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -126,9 +126,15 @@
         <DacpacFile>%(_ResolvedPackageReference.PhysicalLocation)/tools/%(Identity).dacpac</DacpacFile>
         <DatabaseVariableLiteralValue>%(PackageReference.DatabaseVariableLiteralValue)</DatabaseVariableLiteralValue>
       </_ResolvedPackageReference>
+
+      <!-- Resolve all project references to their physical location -->
+      <_ResolvedProjectReference Include="%(_ResolvedProjectReferencePaths.Identity)">
+        <DacpacFile>%(_ResolvedProjectReferencePaths.Identity)</DacpacFile>
```
ErikEJ, line 132: Interesting, I have been struggling with how to do this with code... Inside VS extension.

ErikEJ (approved):

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jmezach: ;) Took some binary logging to figure out how this works. There are some [docs](https://github.com/dotnet/msbuild/blob/master/documentation/ProjectReference-Protocol.md) on how it works if you're interested.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -80,7 +80,7 @@
   <!--
     The GetTargetPathWithTargetPlatformMoniker target uses a BeforeTargets so the only way to disable it is to override it with an empty target.
   -->
-  <Target Name="GetTargetPathWithTargetPlatformMoniker" />
+  <!--<Target Name="GetTargetPathWithTargetPlatformMoniker" />-->
```
jeffrosenberg, line 83: nitpick: why comment this out, rather than remove it entirely?

jeffrosenberg (approved):
One small comment, but it's just a tiny thing. Looks good!

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jmezach: Yes, I agree. It commented it out to try if that resolved my issue, but forgot to remove it after that. Will do that now.

Issue #74, opened 2020-10-04
Support fast up-to-date check


jmezach:
Looks like the up-to-date check is actually a Visual Studio feature that is separate from the MSBuild logic. I found [this document](https://github.com/dotnet/project-system/blob/master/docs/up-to-date-check.md) that describes this feature. We might be able to implement that, although I'm not sure if this will solve the problem at hand.

_Originally posted by @jmezach in https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/63#issuecomment-702880719_

Issue #63, opened 2020-09-22
[Question] Is there a way to set up the project to only build if source files change?


kevcrooks:
At the moment, if I set up the project then run a build it will say "1 succeeded".

If I run the build again without changing any files, I'd like the build to be skipped (i.e. "1 up-to-date" instead of "1 succeeded" in Visual Studio)

Is there a way to do this, or is it not possible, because we always need to reproduce the `dacpac` file?

jeffrosenberg:
We do currently support incremental builds. I'm not sure about what controls the specific verbiage coming out of MSBuild, but if no files are changed, the project isn't re-built.

jmezach:
@kevcrooks It should not recreate the `dacpac` file on every build, but I do believe that Visual Studio still reports `1 succeeded` instead of `1 up-to-date`. I'm not entirely sure why that is. Perhaps the way we've implemented it doesn't match with the expectations that Visual Studio has.

Is this a problem for you? If so I would like to understand why. We shouldn't be "waisting" resources.

kevcrooks:
Thank you, I'll check this, the main reason I'd like this is so I can add a `dotnet publish` to our Database as a post-build step (before my dependent projects), but I think it was getting into a build loop because the `dotnet publish` was doing another build.

I may not be doing this the best way, but I noticed that I was previously using `1.6.0`, so I'll see if this works now I'm on `1.8.1` - please let me know if there's a better way for me to go about a `publish` as part of the build though

kevcrooks:
I've just checked and I'm seeing the same on `1.8.1`.

The main thing I'd like is to be able to deploy the Database as part of a build - but only if the build has changed (so I'm not needing to deploy every time). At the moment I've tried adding:

`dotnet publish --no-build --no-restore --no-dependencies`

as a post-build step, but this is triggering a recurring set of builds.

My backup plan is to create a separate project that only calls `dotnet publish` as a build step, and make all my other projects dependent on that one instead, but I wanted to check if there's a nicer way to do this? And also this has the issue of deploying the Database every time, even if the files haven't changed.

jeffrosenberg:
[...]

PR #75, opened 2020-10-04
Fix up-to-date-check issue

jmezach:
We discovered an issue with the Fast Up-to-date check feature of Visual Studio. It seemed to be looking for debug symbols, but they are never produced by an MSBuild.Sdk.SqlProj project. So we made that explicit now, which seems to fix the issue with fast-up-to-date check.

Fixes: #74


jmezach:
@jeffrosenberg @ErikEJ Turns out that the fast up-to-date check done by Visual Studio didn't work for project using this SDK due to it expecting a `.pdb` file alongside the `.dacpac`. Setting this property it no longer looks for that. I think this a good fix to have regardless of whether it fixes #63, agreed?

ErikEJ (approved):

jeffrosenberg (approved):

Issue #70, opened 2020-10-01
Deploying to older version of SQL Server error message is not clear


brettrowberry:
I have a SQL Server 2016 database server.
I have a project file that includes `<SqlServerVersion>Sql130</SqlServerVersion>`.
`dotnet publish` works great!

`Sql120` also works. However, with `Sql140`, publishing fails with this error message. It's not obvious that it is due to a version mismatch.

```
EXEC : error : Deployment of database 'Prediction' failed: An error occurred during deployment plan generation. Deployment cannot continue. [C:\dev\Prediction\Prediction.fsproj]
C:\Users\User\.nuget\packages\msbuild.sdk.sqlproj\1.8.1\Sdk\Sdk.targets(220,5): error MSB3073: The command "dotnet "C:\Users\User\.nuget\packages\msbuild.sdk.sqlproj\1.8.1\Sdk\../tools/netcoreapp3.1/DacpacTool.dll" deploy -i "C:\dev\Prediction\bin\Debug\netstandard2.0\/Prediction.dacpac" -tsn "localhost\MyServer" -tdn "Prediction"   -p IncludeCompositeObjects=true  " exited with code -1. [C:\dev\Prediction\Prediction.fsproj]
```

jmezach:
Yes I noticed that the error messages being reported when the deployment is failing doesn't give the user much to go on. In fact when I was recording the dotnetFlix episode mentioned in the README I experienced this exact same problem.

I'm not sure what we can do about this. I think we'll need to do some digging to figure out if we can get more details about what's going wrong. I might have some time this weekend.

PR #76, opened 2020-10-04
Improved error handling for publish pipeline

jmezach:
As mentioned in #70 the output for the publish pipeline isn't ideal which leads to users having to figure out by themselves what's going wrong. This PR changes the way we handle exceptions by looking for an InnerException and writing the message from that, rather than the `DacServicesException` which tells us nothing more than that the deployment failed.

I've also hooked up the `Message` event so that we get some more output from the deployment. Essentially you now get the same output as you would get when running `SqlPackage.exe` which can be useful to diagnose problems.

Fixes: #70

### DIFF
--- a/src/DacpacTool/PackageDeployer.cs
+++ b/src/DacpacTool/PackageDeployer.cs
@@ -92,12 +92,33 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             try
             {
                 var services = new DacServices(ConnectionStringBuilder.ConnectionString);
+                services.Message += HandleDacServicesMessage;
                 services.Deploy(Package, targetDatabaseName, true, DeployOptions);
                 _console.WriteLine($"Successfully deployed database '{targetDatabaseName}'");
             }
-            catch (Exception ex)
+            catch (DacServicesException ex)
             {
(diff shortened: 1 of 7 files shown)


src/DacpacTool/PackageDeployer.cs
```diff
@@ -92,12 +92,33 @@ public void Deploy(string targetDatabaseName)
             try
             {
                 var services = new DacServices(ConnectionStringBuilder.ConnectionString);
+                services.Message += HandleDacServicesMessage;
                 services.Deploy(Package, targetDatabaseName, true, DeployOptions);
                 _console.WriteLine($"Successfully deployed database '{targetDatabaseName}'");
             }
-            catch (Exception ex)
+            catch (DacServicesException ex)
```
ErikEJ, line 99: Other exception types?

ErikEJ (changes requested):

src/DacpacTool/PackageDeployer.cs
jmezach: Uhm, yes, of course. Not sure what I was thinking ;). Added a catch for the generic Exception now.

ErikEJ (approved):

jeffrosenberg (approved):

Issue #68, opened 2020-09-30
Dotnet publish does not work when DACPAC output name is changed


RVink:
As described in #65 it's possible to change the name of the DACPAC output by setting the `AssemblyName` in the project file. However changing the output name breaks the `dotnet publish` command. This is because the publish command tries to use the DACPAC file based on the project name.

`C:\Users\rvink\.nuget\packages\msbuild.sdk.sqlproj\1.7.0\Sdk\Sdk.targets(199,5): error MSB3073: The command "dotnet "C:\Users\rvink\.nuget\packages\msbuild.sdk.sqlproj\1.7.0\Sdk\../tools/netcoreapp3.1/DacpacTool.dll" deploy -i "C:\Users\rvink\source\repos\TestProject\TestDatabase\bin\Debug\netstandard2.0\/TestDatabase.dacpac" -tsn "(local)" -tdn "RandomDatabaseName"   -p CreateNewDatabase=False -p IncludeCompositeObjects=true  " exited with code 1. [C:\Users\rvink\source\repos\TestProject\TestDatabase\TestDatabase.csproj]`

This is my project file:
```
<Project Sdk="MSBuild.Sdk.SqlProj/1.7.0">
    <PropertyGroup>
        <TargetFramework>netstandard2.0</TargetFramework>
        <SqlServerVersion>Sql140</SqlServerVersion>
	<AssemblyName>RandomDatabaseName</AssemblyName>
	<TargetDatabaseName>RandomDatabaseName</TargetDatabaseName>
        <!-- For additional properties that can be set here, please refer to https://github.com/rr-wfm/MSBuild.Sdk.SqlProj#model-properties -->
    </PropertyGroup>

    <PropertyGroup>
        <!-- Refer to https://github.com/rr-wfm/MSBuild.Sdk.SqlProj#publishing-support for supported publishing options -->
    </PropertyGroup>
</Project>
```

brettrowberry:
I just observed the same thing. I guess I tried building but not publishing when I added `AssemblyName`. 

I'm documenting my thoughts here.

`dotnet build` works when the following are defined:
- `AssemblyName`
- `TargetDatabaseName`
- `AssemblyName` and `TargetDatabaseName`
- neither is defined

`dotnet publish` works when the following are defined:
- `TargetDatabaseName`
- neither is defined

`dotnet publish` does not work when the following are defined:
- `AssemblyName`
[...]

PR #77, opened 2020-10-04
Allow specifying an alternative file name for the .dacpac

jmezach:
We should allow users to allow to specify a different name for the `.dacpac` file if they choose to do so. This could already be achieved by specifying `<AssemblyName>` in the `.csproj` but when specifying an alternative name it would break `dotnet publish` since it would still look for the `.dacpac` using the projects name. This has now been fixed.

While working on this I also discovered the `<TargetName>` property which can also be set to control the name of the output file, which probably makes a little bit more sense than setting '`<AssemblyName>'. I'll make sure to add that to the docs before completing this PR.

It is probably also worth noting that we currently expect the `.dacpac` to have the same name as the NuGet package when used as a '<PackageReference>`. I'll make sure that I add that to the documentation as well.

Fixes: #68

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -33,8 +33,6 @@
   </ItemGroup>
 
   <PropertyGroup>
-    <TargetExt>.dacpac</TargetExt>
-    <TargetPath>$(TargetDir)/$(MSBuildProjectName)$(TargetExt)</TargetPath>
     <!--We won't be generating debug symbols here -->
     <_DebugSymbolsProduced>false</_DebugSymbolsProduced>
     <CoreBuildDependsOn>
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Setup .NET Core
       uses: coderpatros/setup-dotnet@sxs
       with:
-        dotnet-version: 2.1.809, 3.1.401, 5.0.100-preview.8.20417.9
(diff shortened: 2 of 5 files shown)


ErikEJ (approved):

jmezach:
@jeffrosenberg Could you give this a review as well? Then I can merge this and release this, together with the other PR's as 1.9.0. They are all pretty minor changes, except for the project reference support I guess, so it made sense to bundle them into a single release.

I've also added the documentation stuff I talked about it and bumped the pipeline to use .NET 5.0 RC1.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -33,8 +33,6 @@
   </ItemGroup>
 
   <PropertyGroup>
-    <TargetExt>.dacpac</TargetExt>
-    <TargetPath>$(TargetDir)/$(MSBuildProjectName)$(TargetExt)</TargetPath>
```
jeffrosenberg, line 37: Just to clarify, this is the functional change? Once we remove the explicit `TargetPath`, MSBuild incorporates `TargetName` when it builds the path?

jmezach: @jeffrosenberg Well, it's a little bit more complicated than that. I had the `<TargetPath>` set here because otherwise I wouldn't get a `.dacpac` as the output, but a `.dll` even though I had set `<TargetExt>`. I now realized that we needed to have the `<TargetExt>` property set in the `Sdk.props` file rather than `Sdk.targets`. The props file is imported first, so setting the `<TargetExt>` there makes sure that by the time the `Sdk.targets` file is imported the `<TargetPath>` property is set to the correct value automatically.

jeffrosenberg (approved):

PR #66, opened 2020-09-29
Explanation of `AssemblyName`

brettrowberry:
Explanation of `AssemblyName`

Resolves #65

PR #78, opened 2020-10-05
AssemblyName is now TargetName

brettrowberry:
I'm not sure if this line should be edited or removed.

The problem was introduced by me in #66


jmezach (approved):
LGTM. I think we want to push people to use `<TargetName>` as the preferred method although `<AssemblyName>` will also work. I just feel `<TargetName>` makes more sense in this case.

brettrowberry:
I didn’t test. In 1.8.1 AssemblyName made publish fail. Does it work in 1.9.0?

jmezach:
Yes, with 1.9.0 publish works with both `<TargetName>` and `<AssemblyName>` as expected.

jeffrosenberg (approved):

jmezach:
@brettrowberry Thanks for your contribution.

PR #85, opened 2020-11-04
Update README.md

brettrowberry:
Typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Similar to package references you can also reference another project by using a 
 </Project>
 ```
 
-This will ensure that `MyOtherProject` is build first and the resulting `.dacpac` will be referenced by this project. This means you can use the objects defined in the other project within the scope of this project. If the other project is representing an entirely different database you can also use `DatabaseVariableLiteralValue` attribute on the `ProjectReference` similar to `PackageReference`:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,7 +180,7 @@ Similar to package references you can also reference another project by using a
 </Project>
 ```
 
-This will ensure that `MyOtherProject` is build first and the resulting `.dacpac` will be reference by this project. This means you can use the objects defined in the other project within the scope of this project. If the other project is representing an entirely different database you can also use `DatabaseVariableLiteralValue` attribute on the `ProjectReference` similar to `PackageReference`:
+This will ensure that `MyOtherProject` is build first and the resulting `.dacpac` will be referenced by this project. This means you can use the objects defined in the other project within the scope of this project. If the other project is representing an entirely different database you can also use `DatabaseVariableLiteralValue` attribute on the `ProjectReference` similar to `PackageReference`:
```
ErikEJ, line 183: "is built"

jmezach (approved):

README.md
brettrowberry: Updated

ErikEJ (approved):

jmezach:
@brettrowberry Thanks for your contribution!

Issue #97, opened 2020-11-12
Support for specifying a non-default port number


jmezach:
As discussed in #93 there is a way for users to specify a non-default port number when publishing a `.dacpac` to a SQL Server instance. Unfortunately it requires ugly escaping tricks to make it work properly. Ideally we would provide a `TargetPort` property that can be set at deployment time for users that need it, while we default to not having it specified.

jmezach:
Fixed in #100

PR #100, opened 2020-11-20
Allow users to specify an explicit port for publishing (#97)

jeffrosenberg:
This PR creates a new `UseTargetServerAndPort` method, called only if the user passes in a port. If a port isn't specified, rather than using a default port of 1433, we just fall back to the existing method of not including the port in the connection string at all.

### DIFF
--- a/src/DacpacTool/Program.cs
+++ b/src/DacpacTool/Program.cs
@@ -47,7 +47,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             {
                 new Option<FileInfo>(new string[] { "--input", "-i" }, "Path to the .dacpac package to deploy"),
                 new Option<string>(new string[] { "--targetServerName", "-tsn" }, "Name of the server to deploy the package to"),
-                new Option<string>(new string[] { "--targetPort", "-tprt" }, "Port number to connect on (leave blank for default)"),
+                new Option<int>(new string[] { "--targetPort", "-tprt" }, "Port number to connect on (leave blank for default)"),
                 new Option<string>(new string[] { "--targetDatabaseName", "-tdn" }, "Name of the database to deploy the package to"),
                 new Option<string>(new string[] { "--targetUser", "-tu" }, "Username used to connect to the target server, using SQL Server authentication"),
                 new Option<string>(new string[] { "--targetPassword", "-tp" }, "Password used to connect to the target server, using SQL Server authentication"),
@@ -198,9 +198,9 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                     }
                 }
 
-                if (options.TargetPort != null && Int32.TryParse(options.TargetPort, out int targetPort))
+                if (options.TargetPort.HasValue)
                 {
-                    deployer.UseTargetServerAndPort(options.TargetServerName, targetPort);
+                    deployer.UseTargetServerAndPort(options.TargetServerName, options.TargetPort.Value);
                 }
                 else
                 {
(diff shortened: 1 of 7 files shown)


jeffrosenberg:
@jmezach the CI build is failing on the "Setup .NET Core" step, which this PR doesn't touch at all. Any idea what's going on there? I tried rerunning in case it was just a temporary environment glitch, but it didn't help. I can try rerunning it again in a few hours.

jmezach:
@jeffrosenberg Looks like something has changed on the GitHub Actions side of things. Due to a security vulnerability they have disabled certain features which we are apparently using. I also see that we're still using a custom task to install the .NET SDK on the build agent. We're also still using RC2 of the .NET 5.0 SDK.

Guess we'll have to do some maintenance work here. I'll get on to that when I have some time ;).

jmezach:
So I've modified the workflow to use the default task to install .NET SDK's, which now supports installing multiple versions and that at least gets us past that step in the pipeline. But now it seems to be failing on something that's in the `sqltoolsservice` project that we took a dependency on to fix the SQLCMD file merging. I'll try to update that to a newer version to see if that fixes things.

jmezach:
After updating a whole bunch of stuff everything seems to work now. I still had a failing job, but that was due to a connection reset so I'm guessing that it is a temporary issue.

jeffrosenberg:
@jmezach thanks, looks like everything passes now! Sorry my simple PR led to you having to make 6 commits 🤣

jmezach:
@jeffrosenberg No problem. We'd have to fix this eventually anyway. It is just too bad that there's no way to test out these changes locally ;).

ErikEJ (approved):

src/DacpacTool/Program.cs
```diff
@@ -47,6 +47,7 @@ static async Task<int> Main(string[] args)
             {
                 new Option<FileInfo>(new string[] { "--input", "-i" }, "Path to the .dacpac package to deploy"),
                 new Option<string>(new string[] { "--targetServerName", "-tsn" }, "Name of the server to deploy the package to"),
+                new Option<string>(new string[] { "--targetPort", "-tprt" }, "Port number to connect on (leave blank for default)"),
```
jmezach, line 50: I think we can have an Option<int> here so that we don't have to do any parsing ourselves.

jmezach (approved):
LGTM. Only 1 comment.

src/DacpacTool/Program.cs
jeffrosenberg: I've pushed a change, but C# isn't my primary language, so let me know if it's not the idiomatic way to do things 🙂

src/DacpacTool/Program.cs
```diff
@@ -197,7 +198,14 @@ private static int DeployDacpac(DeployOptions options)
                     }
                 }
 
-                deployer.UseTargetServer(options.TargetServerName);
+                if (options.TargetPort != null)
```
jmezach, line 201: Perhaps more idiomatic C# would be to do `if (options.TargetPort.HasValue)` here ;).

jeffrosenberg: Made that change, I believe we should be good to merge this?

jmezach:
Yes I'm okay with merging this. Feels like a new feature so do we want to ship this as 1.10.0?

jmezach:
@jeffrosenberg Thanks for your contribution!

jmezach:
@jeffrosenberg Would you mind adding this to the README as well? I noticed we didn't change that as part of this PR.

PR #111, opened 2021-01-16
Allow running pre/post deployment scripts for referenced packages

### LINKED ISSUE #110: Posibility to run post-deployment scripts of dependencies
Can it be possible to run the pre- and post deployment scripts of your dependencies too?

jmezach:
When referencing other database packages, either by using a `ProjectReference` or a `PackageReference`, the pre- and/or post-deployment scripts of those dependencies are not run during deployment. While this is understandable in a deployment pipeline when using `SqlPackage.exe` it can be desirable when developers need to update their local development databases. For example, when the post-deployment script inserts some well-known data into certain tables, a developer might expect to have that data in his local development database. In this PR I've added opt-in support to allow for this.

Fixes #110

### DIFF
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -25,10 +25,7 @@
     <EmbeddedResource Include="$(ManagedBatchParserPath)/Localization/sr.resx" LogicalName="Microsoft.SqlTools.ManagedBatchParser.Localization.SR.resources" />
     <None Include="$(ManagedBatchParserPath)/Localization/sr.strings" />
     <Compile Include="$(ManagedBatchParserPath)/BatchParser/**/*.cs" />
-    <Compile Remove="$(ManagedBatchParserPath)/BatchParser/ExecutionEngineCode/ExecutionEngine.cs" />
-    <Compile Remove="$(ManagedBatchParserPath)/BatchParser/ExecutionEngineCode/Batch.cs" />
-    <Compile Remove="$(ManagedBatchParserPath)/BatchParser/ExecutionEngineCode/BatchParser*Args.cs" />
-    <Compile Remove="$(ManagedBatchParserPath)/BatchParser/BatchParserWrapper.cs" />
+    <Compile Include="$(ManagedBatchParserPath)/ReliableConnection/**/*.cs" />
   </ItemGroup>
 
 </Project>
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -12,14 +12,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SqlServer.DACFX" Version="150.4826.1" />
-    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20427.1" />
-    <PackageReference Include="System.Security.Permissions" Version="4.7.0" />
+    <PackageReference Include="Microsoft.SqlServer.DACFX" Version="150.4897.1" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20574.7" />
+    <PackageReference Include="System.Security.Permissions" Version="5.0.0" />
   </ItemGroup>
 
   <!-- References to Microsoft.SqlTools.ManagedBatchParser -->
   <ItemGroup>
-    <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="161.41011.9" />
+    <PackageReference Include="Microsoft.SqlServer.SqlManagementObjects" Version="161.46041.41" />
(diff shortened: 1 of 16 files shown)


jmezach:
@jeffrosenberg @ErikEJ If you could have a look through the changes that would be great. I'll ask my team to test-drive a pre-release version of this before completing this PR.

src/DacpacTool/DacpacTool.csproj
```diff
@@ -12,23 +12,20 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SqlServer.DACFX" Version="150.4826.1" />
-    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20427.1" />
-    <PackageReference Include="System.Security.Permissions" Version="4.7.0" />
+    <PackageReference Include="Microsoft.SqlServer.DACFX" Version="150.4897.1" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20574.7" />
+    <PackageReference Include="System.Security.Permissions" Version="5.0.0" />
```
ErikEJ, line 17: Is this update required?

src/DacpacTool/Extensions.cs
```diff
@@ -27,6 +113,29 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
...
+
+            MethodInfo getMetadataMethod = null;
+            foreach (var reference in references)
+            {
+                if (getMetadataMethod == null)
+                {
+                    getMetadataMethod = reference.GetType().GetMethod("GetMetadata", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(string) }, null);
+                }
+
+                var fileName = (string)getMetadataMethod.Invoke(reference, new object[] { "FileName" });
```
ErikEJ, line 132: Add null check?

ErikEJ (changes requested):

src/DacpacTool/DacpacTool.csproj
jmezach: Not necessarily no, but I'd rather stay up to date with our dependencies. Do you have a concern about these updates that I'm not aware of?

src/DacpacTool/Extensions.cs
jmezach: Do you mean on the result of `GetMethod()` or on the result if `Invoke()`?

src/DacpacTool/DacpacTool.csproj
ErikEJ: I think it can be removed, actually

jmezach: You mean the System.Security.Permissions package?

src/DacpacTool/Extensions.cs
ErikEJ: I meant before doing a result.Add()

src/DacpacTool/DacpacTool.csproj
ErikEJ: Yes, the DacFX package now brings in the correct, newer version

jmezach: Yup, seems to work just fine.

src/DacpacTool/Extensions.cs
jmezach: Ah, yes, that makes sense. Added that now.

ErikEJ (approved):

src/DacpacTool/PackageBuilder.cs
```diff
@@ -1,12 +1,11 @@
 ﻿using System;
 using System.IO;
+using System.IO.Packaging;
```
jeffrosenberg, line 3: I don't understand this... the commit message says "respond to breaking changes", so is this change related to some upstream dependency?

jeffrosenberg (approved):
LGTM, just one thing I didn't understand that I would like to get some context for

src/DacpacTool/PackageBuilder.cs
ErikEJ: I think the DacFX package no longer includes a copy of the Packaging API

jmezach: Yeah, I think that's what happend as well. After upgrading the DacFX package the  `Microsoft.Data.Tools.Schema.Sql.Packaging` no longer existed. So I figured I'd just replace it with the built-in `System.IO.Packaging` API to see if that works. It turned out to be basically a drop-in replacement. I'm assuming that when they originally built DacFX that the Packaging API didn't exist in .NET yet. But now it does so they decided to replace their own implementation with the one from the framework.

ErikEJ: I had the exact same experience when updating my .dacpac reverse engineering library.

jmezach:
I've added an additional note to the README that `DefaultValue` isn't actually used prior to this version. I though I'd make that clear in the docs. I've heard back from my team that this is working well for our scenario, so as soon as CI completes I'll merge this PR and release this as 1.11.0.

Issue #114, opened 2021-01-21
:r relative path not working on linux


PyroSA:
Version: MSBuild.Sdk.SqlProj/1.11.1

We're using mcr.microsoft.com/dotnet/core/sdk:3.1-buster on our docker-based build.

**Expectation:** 
:r scripts in a relative path should be included and and run in the post-deploy script.

**Actual:**
Error - file not found.

**Detail:**
The container is Debian 10 based - Linux.

When running dotnet build on there for the dacpac project, we have an issue with paths.

We have our data insertions separated into scripts files under a data folder.
These are then included in the post-build script as:
```sql
:r .\Data\InsertScript1.sql
:r .\Data\InsertScript2.sql
```

But when we run the build from a Dockerfile, we get the following output:
```
Microsoft (R) Build Engine version 16.7.2+b60ddb6f4 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored /src/DB.build.csproj (in 222 ms).
  Unhandled exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
   ---> System.IO.FileNotFoundException: Could not find file '/src/.\Data\InsertScript1.sql'.
  File name: '/src/.\Data\InsertScript1.sql'
     at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
     at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
     at System.IO.FileStream.OpenHandle(FileMode mode, FileShare share, FileOptions options)
     at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
     at System.IO.StreamReader.ValidateArgsAndOpenPath(String path, Encoding encoding, Int32 bufferSize)
     at System.IO.StreamReader..ctor(String path)
     at MSBuild.Sdk.SqlProj.DacpacTool.ScriptParser.Microsoft.SqlTools.ServiceLayer.BatchParser.ICommandHandler.Include(TextBlock includeBlock, TextReader& stream, String& newFilename) in /home/runner/work/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj/src/DacpacTool/ScriptParser.cs:line 64
[...]

PR #116, opened 2021-01-26
Exclude scripts from bin/ and obj/ directories, add additional test case per discussion on #114

jeffrosenberg:
Nothing major here -- just a few quick things that came up in the discussion of #114:
1. Edited Sdk.props to use forward-slashes instead of backslashes
2. Edited Sdk.props to exclude content from `bin/` and `obj`
3. Added an additional test to `ScriptParserTests.cs` to double-check the issue reported in #114, and decided that since I had already added it, I may as well keep it in there

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
@@ -37,11 +37,13 @@
   <Target Name="CompileDesignTime" />
 
   <ItemGroup>
-    <Content Include="**\*.sql">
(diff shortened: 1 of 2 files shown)


ErikEJ (approved):

ErikEJ:
LGTM!

jmezach (approved):
LGTM

jmezach:
I would suggest releasing this as a patch. Agreed? @jeffrosenberg @ErikEJ

ErikEJ:
Agreed! 
See you on Wed, btw!

PR #129, opened 2021-04-30
Adding support SuppressTSqlWarnings for particular files

McHarny:
Added functionality similar to `sqlproj` projects where you can specify some warnings to be suppressed for specific files.
So you can add `SuppressTSqlWarnings` element under `Content` element:
```xml
<Project Sdk="MSBuild.Sdk.SqlProj/1.12.0">
    <PropertyGroup>
        ...
    </PropertyGroup>

    <ItemGroup>
        <Content Include="Procedures\csp_Test.sql">
            <SuppressTSqlWarnings>71502</SuppressTSqlWarnings>
        </Content>
    </ItemGroup>
</Project>
```
During build process there are creation of file `InputFiles.txt` where every line contains path to file for build. I also make second file `WarningsSuppression.txt` where every line contains path to file , `|` delimeter, and contents of SuppressTSqlWarnings metadata for this file. As of format logic it must be comma-separated list of warning codes to suppress.
For example:
```
Procedures\csp_Test.sql|71502
Procedures\csp_Test1.sql|71502,71558
```
Path to this file pass as the parameter to `dacpactool` and during validation `PackageBuilder` will suppress this warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,32 @@ Refer to the [documentation](https://docs.microsoft.com/dotnet/api/microsoft.sql
 
 **Note:** If you are replacing an existing `.sqlproj` be sure to copy over any of these properties into the new project file.
 
+## Model compiler options
+Like an `.sqlproj` projects  `MSBuild.Sdk.SqlProj` supports special `msbuild` options how to process T-SQL compiling warnings.
+Treating warnings as errors can be optionally enabled by adding a property `TreatTSqlWarningsAsErrors` to the project file as in the below example:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/1.11.0">
+    <PropertyGroup>
+        <TreatTSqlWarningsAsErrors>True</TreatTSqlWarningsAsErrors>
+        ...
+    </PropertyGroup>
+</Project>
+```
+
+Suppression of some warnings from treating as errors can be done by adding comma-separated (or semicolon) list of warning codes to `SuppressTSqlWarnings` property in the project file:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/1.11.0">
+    <PropertyGroup>
+        <SuppressTSqlWarnings>71558,71502</SuppressTSqlWarnings>
+        <TreatTSqlWarningsAsErrors>True</TreatTSqlWarningsAsErrors>
+        ...
+    </PropertyGroup>
+</Project>
+```
+
+
 ## Pre- and post deployment scripts
 Support for pre- and post deployment scripts has been added in version 1.1.0. These scripts will be automatically executed when deploying the `.dacpac` to SQL Server.
 
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ You can suppress warnings for a specific file by adding `SuppressTSqlWarnings` f
     </ItemGroup>
 </Project>
 ```
-
+> Note: Warnings suppressed at the project level are always applied to every file in the project, regardless of what is configured at the file level.
 
 ## Pre- and post deployment scripts
 Support for pre- and post deployment scripts has been added in version 1.1.0. These scripts will be automatically executed when deploying the `.dacpac` to SQL Server.
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ dotnet new sqlproj -s Sql130
 You should now have a project file with the following contents:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/1.11.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/1.12.0">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
         <SqlServerVersion>Sql130</SqlServerVersion>
@@ -92,7 +92,7 @@ If you already have a SSDT (.sqlproj) project in your solution, you can keep tha
 There are a lot of properties that can be set on the model in the resulting `.dacpac` file which can be influenced by setting those properties in the project file using the same name. For example, the snippet below sets the `RecoveryMode` property to `Simple`:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/1.11.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/1.12.0">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
         <RecoveryMode>Simple</RecoveryMode>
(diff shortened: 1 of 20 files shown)


src/DacpacTool/Program.cs
```diff
@@ -138,6 +139,23 @@ private static int BuildDacpac(BuildOptions options)
                 packageBuilder.AddWarningsToSuppress(options.SuppressWarnings);
             }
 
+            // Add warnings suppressions for particular files through $Project.WarningsSuppression.txt
+            if (options.SuppressWarningsListFile != null)
+            {
+                if (options.SuppressWarningsListFile.Exists)
```
jmezach, line 145: Perhaps this logic should move into the `PackageBuilder` as well. `Program` is merely responsible for parsing command line arguments and passing them to the `PackageBuilder `.

README.md
```diff
@@ -131,6 +131,23 @@ To suppress specific warnings from being treated as errors, add a comma-separate
 </Project>
 ```
 
+You can suppress warnings for a specific file by adding `SuppressTSqlWarnings` for the this file:
```
jmezach, line 134: Remove 'the' here

README.md
```diff
@@ -131,6 +131,23 @@ To suppress specific warnings from being treated as errors, add a comma-separate
...
+    </PropertyGroup>
+
+    <ItemGroup>
+        <Content Include="Procedures\csp_Test.sql">
+            <SuppressTSqlWarnings>71502</SuppressTSqlWarnings>
+        </Content>
+    </ItemGroup>
+</Project>
+```
+
```
jmezach, line 149: We should probably explain how this works in combination with the project-level `SuppressTSqlWarnings`. Looking at the code it seems that if a particular warning is suppressed at the project level it will always be suppressed regardless of what is configured at the file level. I think that behaviour is fine, but we should at least explain that in the docs.

jmezach (changes requested):
Thanks again @McHarny for your contribution. I've made a couple of comments. I also think this could use some unit tests to validate this functionality.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -187,7 +187,18 @@
     <WriteLinesToFile
       File="$(IntermediateOutputPath)$(MSBuildProjectName).InputFiles.txt"
       Overwrite="true"
-      Lines="@(Content)" />
+      Lines="@(Content->Distinct())" />
```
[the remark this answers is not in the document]
McHarny: In `Sdk.props` at line `<Content Include="**/*.sql">` we Include **all** .sql files as Content to compile.
When we again specify 
```xml
<ItemGroup>
    <Content Include="Procedures\csp_Test.sql">
        <SuppressTSqlWarnings>71502</SuppressTSqlWarnings>
    </Content>
</ItemGroup>
```
We add this file to `Content` list again. So it contains two "links" for that file. To avoid this we must at first remove reference to file from `Content` list, and then include it. Like below.
```xml
<ItemGroup>
    <Content Remove="Procedures\csp_Test.sql">
    <Content Include="Procedures\csp_Test.sql">
        <SuppressTSqlWarnings>71502</SuppressTSqlWarnings>
    </Content>
</ItemGroup>
```

I think that this is too hard to remeber to specify "Remove" for every file. I prefer to simply copy section from `sqlproj` file
and remove duplicates from `Content` list

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
```diff
@@ -39,6 +39,7 @@
   <ItemGroup>
     <Content Include="**/*.sql">
       <Pack>False</Pack>
+      <SuppressTSqlWarnings></SuppressTSqlWarnings>
```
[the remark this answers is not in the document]
McHarny: Yes, it is nessesary to add `SuppressTSqlWarnings` metadata for all default `Content` items. Because if we doesn't add it, and make reference to it, MSBuild fail to process

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -187,7 +187,18 @@
...
+
+    <!-- Write the list of input files whith warnings suppression to a file to be consumed by the dacpac tool -->
+    <ItemGroup>
+      <!-- Build a list of particular files with T-SQL warning suppression -->
+      <WarningsSuppressionFiles Include="@(Content->'%(Identity)|%(SuppressTSqlWarnings)')" Condition="'%(Content.SuppressTSqlWarnings)' != ''"/>
```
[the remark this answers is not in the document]
McHarny: Update main comment

README.md
McHarny: Sorry but I'm afraid that my English will not give me the opportunity to write a competent explanation of this behavior. this behavior is the same as at standard SSDT project.

src/DacpacTool/Program.cs
McHarny: I make this by analogy with how it is processed standard `InputFiles.txt`. `Program` parse this file and pass to `PackageBuilder` particular paths.

McHarny:
@jmezach I add Unit test, correct `README`, and make answers for your comments.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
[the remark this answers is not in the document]
jeffrosenberg: That explanation makes sense to me! 👍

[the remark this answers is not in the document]
jmezach: I think you should be able to do:

```xml
<ItemGroup>
    <Content Update="Procedures\csp_Test.sql">
        <SuppressTSqlWarnings>71502</SuppressTSqlWarnings>
    </Content>
</ItemGroup>
```

which would avoid the issue. But I guess it can't hurt to do the Distinct as well :).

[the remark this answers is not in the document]
McHarny: Yea you can use `Update` clause, but because all editing make manually in large project with many files, folders etc. It became too difficult to monitor what file you must `Include`, and which you must `Update`

src/DacpacTool/Program.cs
```diff
@@ -138,6 +139,23 @@ private static int BuildDacpac(BuildOptions options)
...
+                {
+                    foreach (var line in File.ReadLines(options.SuppressWarningsListFile.FullName))
+                    {
+                        //Checks if there are suppression warnings list
+                        var parts = line.Split('|', StringSplitOptions.RemoveEmptyEntries);
+                        var warningList = (parts.Length > 1) ? parts[1] : null;
+
+                        FileInfo inputFile = new FileInfo(parts[0]); // Validation occurs in AddInputFile
```
jmezach, line 153: The comment here doesn't match what's actually happening I think.

README.md
jmezach: Perhaps it is sufficient to add the following after the example:

```
> Note: Warnings suppressed at the project level are always applied to every file in the project, regardless of what is configured at the file level.
```

jmezach (approved):
I've added a couple of additional comments, but overall it looks good to me.

src/DacpacTool/Program.cs
McHarny: OK. IMHO:
`PackageBuilder` has method `AddInputFile` - which adds the file to process at build phase, and it take one argument - `path to file`. So `PackageBuilder` has second method `AddFileWarningsToSuppress` which add suppression warnings for file. and it take two arguments - `path to file` & `suppression list`.
So `PackageBuilder` doesn't do any file parsing. It always get complete atomic argument to deal. For example in Unit Test we don't need to generate any sample `WarningsSuppression.txt`, but simple call `AddFileWarningsToSuppress` with `file` and `suppression list` arguments. `PackageBuilder` is absolutely free from knowing from what source we get arguments for his methods.
Also, IMHO, if we want to extract file parsing from `Program.cs`, we need to make middleware class between `Program` and `PackageBuilder`.

src/DacpacTool/Program.cs
jeffrosenberg: Funny how patterns like this develop from a series of small incremental steps... I based the parsing of `InputFiles.txt` off of the code that existed at the time, which accepted multiple `--input` arguments and sent each to `PackageBuilder.AddInputFile()`. Now that we have multiple of these types of input files, though, I tend to agree that this belongs in `PackageBuilder`. I think the way this was developed makes sense according to the patterns in our code right now, but maybe we should refactor in a follow-up to this PR?

test/DacpacTool.Tests/PackageBuilderTests.cs
```diff
@@ -499,6 +499,28 @@ public void ValidateModel_WarningsAsErrorsSuppressGlobal()
...
+            // Add file with warnings suppression
+            var fileInfo = new FileInfo("../../../../TestProjectWithWarningsSuppressForFile/Procedures/csp_Test.sql");
+            packageBuilder.AddInputFile(fileInfo);
+            packageBuilder.AddFileWarningsToSuppress(fileInfo, "71502");
+
+            // Act
+            bool result = packageBuilder.ValidateModel();
+
+            // Assert
+            // Validation mast be true, SQL71502 Warning for file must be suppressed
```
jeffrosenberg, line 520: Typo, "mast" --> "must"

jeffrosenberg (changes requested):
Just one tiny typo I found 🙂

src/DacpacTool/Program.cs
jmezach: Yes, I agree. I guess my main concern is that we don't currently test the parsing logic for the input file since we don't have any tests for Program. So yeah, this definitely needs a bit of refactoring. But I agree with @jeffrosenberg earlier comment that we should probably do that after this is merged in. So let's live this as it is for now and then we can do a follow-up-PR to refactor this.

jeffrosenberg (approved):

jmezach:
@McHarny Thank you again for your contribution. I'll get this released as 1.13.0.

PR #131, opened 2021-05-05
Support of SQLCMD variables in Project & Package references

McHarny:
Standard SSDT projects has ability to reference other database not only by `DatabaseVariableLiteralValue` but also by using `SQLCMD` variables to use in code for reference another database/server. Like this:
```xml
<ProjectReference Include="SomeProject.sqlproj">
      <Name>SomeProject</Name>
      <Project>{c8cef164-7da7-4fb6-9fcf-41bcd3257218}</Project>
      <Private>True</Private>
      <SuppressMissingDependenciesErrors>False</SuppressMissingDependenciesErrors>
      <DatabaseSqlCmdVariable>SomeDatabase</DatabaseSqlCmdVariable>
      <ServerSqlCmdVariable>SomeServer</ServerSqlCmdVariable>
</ProjectReference>
...
<ItemGroup>
  <SqlCmdVariable Include="SomeDatabase">
      <DefaultValue>SomeDatabase</DefaultValue>
      <Value>$(SqlCmdVar__1)</Value>
    </SqlCmdVariable>
    <SqlCmdVariable Include="SomeServer">
      <DefaultValue>SomeServer</DefaultValue>
      <Value>$(SqlCmdVar__2)</Value>
  </SqlCmdVariable>
</ItemGroup>
```
And objects from this project can be reference by `[$(SomeServer)].[$(SomeDatabase)].[<schema>].[<object>]` syntax.

I add similar functionality to SqlProj.
Since the functionality of `DatabaseVariableLiteralValue` is based on attributes, not elements as in `.sqlproj`, I added two new attributes to the `PackageReference` and `ProjectReference` elements with the names similar to `.sqlproj`:
```
DatabaseSqlCmdVariable
ServerSqlCmdVariable
```
Their behavior is completely similar to that of `.sqlproj`. You can specify them in the reference, define appropriate `SQLCMD` variables and use reference in code. You can combine them with `DatabaseVariableLiteralValue` for example:
```xml
<Project Sdk="MSBuild.Sdk.SqlProj/1.12.0">
    <PropertyGroup>
        <TargetFramework>netstandard2.0</TargetFramework>
    </PropertyGroup>

    <ItemGroup>
        <ProjectReference Include="../MyOtherProject/MyOtherProject.csproj" DatabaseVariableLiteralValue="SomeOtherDatabase" />
    </ItemGroup>
    <ItemGroup>
        <ProjectReference Include="../MySecondProject/MySecondProject.csproj" DatabaseSqlCmdVariable="SecondOtherDatabase" ServerSqlCmdVariable="SomeOtherServer" />
    </ItemGroup>
    <ItemGroup>
        <ProjectReference Include="../MyThirdProject/MyThirdProject.csproj" DatabaseVariableLiteralValue="ThirdOtherDatabase" ServerSqlCmdVariable="SomeOtherServer" />
    </ItemGroup>

  <ItemGroup>
    <SqlCmdVariable Include="SecondOtherDatabase">
      <DefaultValue>SecondDatabase</DefaultValue>
      <Value>$(SqlCmdVar__1)</Value>
    </SqlCmdVariable>
    <SqlCmdVariable Include="SomeOtherServer">
      <DefaultValue>OtherServer</DefaultValue>
      <Value>$(SqlCmdVar__2)</Value>
    </SqlCmdVariable>
  </ItemGroup>

</Project>
```
For this example we can reference other objects with those combinations:
```
[SomeOtherDatabase].[<schema>].[<object>]
[$(SomeOtherServer)].[$(SecondOtherDatabase)].[<schema>].[<object>]
[$(SomeOtherServer)].[ThirdOtherDatabase].[<schema>].[<object>]
```

To do this, I changed the format of the `--reference` (`-r`) parameter. Previously, it took the value `<Path>;<DatabaseVariableLiteralValue>`, now it takes the value `<Path>;<External parts>`, where `<External parts>` has the format `dbl=<DatabaseVariableLiteralValue>|dbv=<DatabaseSqlCmdVariable>|srv=<ServerSqlCmdVariable>`

In `Extensions` class I add method `ParseExternalParts` which parses this `<External parts>`. Parsing of this is made similar to that is writed in constructor of `Microsoft.Data.Tools.Schema.Sql.SchemaModel.SqlSchemaCustomSchemaDataWrapper`

Also parsing supports old behavior with simple `<DatabaseVariableLiteralValue>` passing in the parameter.

I add some Unit-tests to `ExtensionsTest` and add test project `TestProjectWithExternalServerReference`

In addition I slightly changed the behavior of the warning processing to make it more similar to the one in `.sqlproj`. Now the message of warnings threated as error wrote as `error`. Suppressed warnings are not appear in output. 
I remove `ModelValidationError` clause from output text for compatibility with some VS extensions which parses output and colorize it based on regex.

I also correct `CanGetModelValidationErrors` test bacause it failed on localized versions of `.NET`.
Also I have problems with `CanParseFileWithIncludesFromRelativePaths` test on some machines with mixing of `\` and `/` in relative paths, so I add  replacing `\` to `/` in target. It doesn't affect the functionality of test

### DIFF
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -6,7 +6,6 @@ using System.Reflection;
 using System.Text;
 using Microsoft.SqlServer.Dac;
 using Microsoft.SqlServer.Dac.Model;
-using Microsoft.SqlServer.TransactSql.ScriptDom;
 using Microsoft.SqlTools.ServiceLayer.BatchParser.ExecutionEngineCode;
 
 namespace MSBuild.Sdk.SqlProj.DacpacTool
@@ -107,10 +106,85 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
 
             if (!string.IsNullOrWhiteSpace(externalParts))
             {
-                setMetadataMethod.Invoke(customData, new object[] { "ExternalParts", Identifier.EncodeIdentifier(externalParts) });
+                var parts = ParseExternalParts(externalParts);
+                if (!string.IsNullOrEmpty(parts))
+                {
+                    setMetadataMethod.Invoke(customData, new object[] {"ExternalParts", parts});
+                }
             }
 
             AddCustomData(dataSchemaModel, customData);
+        }
+
+        public static string ParseExternalParts(string externalParts)
+        {
+            string serverVariableName = null;
+            string databaseVariableName = null;
+            string databaseVariableLiteralValue = null;
+            foreach (var part in externalParts.Split('|', StringSplitOptions.RemoveEmptyEntries))
+            {
+                if (part.Length < 5)
+                    continue;
+
+                var prefix = part.Substring(0, 4);
+                if (string.Equals(prefix, "dbl=", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    databaseVariableLiteralValue = part.Substring(4);
+                }
+                else if (string.Equals(prefix, "dbv=", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    databaseVariableName = part.Substring(4);
+                }
+                else if (string.Equals(prefix, "srv=", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    serverVariableName = part.Substring(4);
+                }
+            }
+
+            if (string.IsNullOrEmpty(serverVariableName) && string.IsNullOrEmpty(databaseVariableName) &&
+                string.IsNullOrEmpty(databaseVariableLiteralValue) && !externalParts.Contains('='))
+            {
+                databaseVariableLiteralValue = externalParts;
+            }
+            
+            string result = string.Empty;
+            
+            if (string.IsNullOrEmpty(serverVariableName) && !string.IsNullOrEmpty(databaseVariableLiteralValue))
+            {
+                result = "[" + databaseVariableLiteralValue + "]";
+            }
+            else if (!string.IsNullOrEmpty(serverVariableName) && !string.IsNullOrEmpty(databaseVariableLiteralValue))
+            {
+                result = string.Concat("[", EnsureIsDelimited(serverVariableName), "].[", databaseVariableLiteralValue, "]");
+            }
+            else if (string.IsNullOrEmpty(serverVariableName) && !string.IsNullOrEmpty(databaseVariableName))
+            {
+                result = "[" + EnsureIsDelimited(databaseVariableName) + "]";
+            }
+            else if (!string.IsNullOrEmpty(serverVariableName) && !string.IsNullOrEmpty(databaseVariableName))
+            {
+                result = string.Concat("[", EnsureIsDelimited(serverVariableName), "].[", EnsureIsDelimited(databaseVariableName), "]");
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        /// Cached method info for FileUtils.EnsureIsDelimited
+        /// </summary>
+        private static MethodInfo _ensureIsDelimitedMethod = null;
+        /// <summary>
+        /// This method found in Microsoft.Data.Tools.Utilities in class FileUtils. because of it is internal we do call through Reflection
+        /// </summary>
+        private static string EnsureIsDelimited(string name)
+        {
+            if (_ensureIsDelimitedMethod == null)
+            {
+                _ensureIsDelimitedMethod = Type.GetType("Microsoft.Data.Tools.Schema.Common.FileUtils, Microsoft.Data.Tools.Utilities")
+                    .GetMethod("EnsureIsDelimited", BindingFlags.InvokeMethod | BindingFlags.Static | BindingFlags.Public, null, new[]{ typeof(string) }, null);
+            }
+
+            return (string)_ensureIsDelimitedMethod.Invoke(null, new object[]{ name });
         }
 
         public static IEnumerable<string> GetReferencedDacPackages(this TSqlModel model)
@@ -118,7 +192,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             var result = new List<string>();
             var dataSchemaModel = GetDataSchemaModel(model);
 
-            var getCustomDataMethod = dataSchemaModel.GetType().GetMethod("GetCustomData", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(string), typeof(string) }, null);
+            var getCustomDataMethod = dataSchemaModel.GetType().GetMethod("GetCustomData", BindingFlags.Public | BindingFlags.Instance, null, new[] { typeof(string), typeof(string) }, null);
             var references = (IEnumerable) getCustomDataMethod.Invoke(dataSchemaModel, new object[] { "Reference", "SqlSchema" });
 
             MethodInfo getMetadataMethod = null;
@@ -126,7 +200,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             {
                 if (getMetadataMethod == null)
                 {
-                    getMetadataMethod = reference.GetType().GetMethod("GetMetadata", BindingFlags.Public | BindingFlags.Instance, null, new Type[] { typeof(string) }, null);
+                    getMetadataMethod = reference.GetType().GetMethod("GetMetadata", BindingFlags.Public | BindingFlags.Instance, null, new[] { typeof(string) }, null);
                 }
 
                 var fileName = (string)getMetadataMethod.Invoke(reference, new object[] { "FileName" });
@@ -170,7 +244,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                     documentProperty = modelValidationError.GetType().GetProperty("Document", BindingFlags.Public | BindingFlags.Instance);
                 }
 
-                var dacModelError = createDacModelErrorMethod.Invoke(service, new object[] { modelValidationError }) as DacModelError;
+                var dacModelError = createDacModelErrorMethod.Invoke(service, new[] { modelValidationError }) as DacModelError;
                 result.Add(new ModelValidationError(dacModelError, documentProperty.GetValue(modelValidationError) as string));
             }
 
@@ -187,7 +261,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         private static void AddCustomData(object dataSchemaModel, object customData)
         {
             var addCustomDataMethod = dataSchemaModel.GetType().GetMethod("AddCustomData", BindingFlags.Public | BindingFlags.Instance);
-            addCustomDataMethod.Invoke(dataSchemaModel, new object[] { customData });
+            addCustomDataMethod.Invoke(dataSchemaModel, new[] { customData });
         }
     }
 }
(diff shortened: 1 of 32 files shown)


src/DacpacTool/Extensions.cs
```diff
@@ -107,26 +106,101 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
...
+                if (!string.IsNullOrEmpty(parts))
+                {
+                    setMetadataMethod.Invoke(customData, new object[] {"ExternalParts", parts});
+                }
             }
 
             AddCustomData(dataSchemaModel, customData);
         }
 
+        public static string ParseExternalParts(string externalParts)
```
jmezach, line 119: I'm wondering if perhaps a regular expression would be a better solution for this parsing logic. It feels fairly complicated.

src/DacpacTool/ModelValidationError.cs
```diff
@@ -72,5 +72,30 @@ public override string ToString()
 
             return stringBuilder.ToString();
         }
+
+        /// <summary>
+        /// Make message for writing to output during build process more similar to the one displayed by standard SSDT build
+        /// </summary>
+        /// <param name="severity">Severity to override depends on TreatTSqlWarningsAsErrors</param>
+        /// <returns></returns>
+        public string GetOutputMessage(ModelErrorSeverity severity)
```
jmezach, line 81: The original goal of the override of `ToString()` was to do exactly this. Perhaps that override is no longer necessary if we add this?

src/DacpacTool/PackageBuilder.cs
```diff
@@ -112,19 +112,26 @@ public bool ValidateModel()
                 if (modelError.Severity == ModelErrorSeverity.Error)
                 {
                     validationErrors++;
+                    Console.WriteLine(modelError.GetOutputMessage(modelError.Severity));
                 }
-                else if (modelError.Severity == ModelErrorSeverity.Warning && TreatTSqlWarningsAsErrors)
+                else if (modelError.Severity == ModelErrorSeverity.Warning)
                 {
                     if (!_suppressedWarnings.Contains(modelError.ErrorCode))
```
jmezach, line 119: This is getting a little messy with multiple nested if's. Perhaps we could fold the first two if's here into a single method call.

test/DacpacTool.Tests/ExtensionsTest.cs
```diff
@@ -73,7 +73,121 @@ public void CanGetModelValidationErrors()
 
             var error = modelValidationErrors.First();
             error.Severity.ShouldBe(Microsoft.SqlServer.Dac.Model.ModelErrorSeverity.Error);
-            error.ToString().ShouldBe("../../../../TestProjectWithErrors/Procedures/csp_Test.sql(2,18):ModelValidationError Error SQL71501: SqlSubroutineParameter: [dbo].[csp_Test].[@p_Parameter] has an unresolved reference to SqlBuiltInType [dbo].[MyCustomType].");
+            error.ErrorCode.ShouldBe(71501);
```
jmezach, line 76: Did we remove the assertion on ToString() for a reason? I think that still has value to avoid regressions in the output we give.

test/DacpacTool.Tests/ScriptParserTests.cs
```diff
@@ -62,7 +62,7 @@ public void CanParseFileWithIncludesFromRelativePaths()
             // Assert
             result.ShouldNotBeNull();
             result.Count().ShouldBe(2);
-            result.ToList().Take(2).Last().ShouldBe($"{TEST_PROJECT_PATH}/Pre-Deployment/./MoreScripts/Script2.sql");
+            result.ToList().Take(2).Last().Replace("\\", "/").ShouldBe($"{TEST_PROJECT_PATH}/Pre-Deployment/./MoreScripts/Script2.sql");
```
jmezach, line 65: Was this change deliberate? This appears to be a platform specific difference, so perhaps we should handle this better?

test/TestProjectWithExternalServerReference/TestProjectWithExternalServerReference.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <SqlServerVersion>Sql150</SqlServerVersion>
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <ProjectReference Include="..\TestProject\TestProject.csproj" DatabaseVariableLiteralValue="SomeDatabase" ServerSqlCmdVariable="SomeServer"/>
+  </ItemGroup>
+
+  <ItemGroup>
+    <SqlCmdVariable Include="SomeDatabase">
```
jmezach, line 14: This variable isn't actually needed right? Perhaps we should remove it then.

jmezach (changes requested):
Once again, thank you for your contribution @McHarny. Overall I think this looks good, just a couple of comments.

src/DacpacTool/ModelValidationError.cs
McHarny: IMHO override of `ToString` is good practice for develop & debug purpose.

test/DacpacTool.Tests/ExtensionsTest.cs
McHarny: If we has localized `.NET` on machine running tests it report error message other than that with which you are compare. So I think that we must check `ModelValidationError` properties to check correct result, instead of check `ToString`

test/DacpacTool.Tests/ScriptParserTests.cs
McHarny: OK. This issue depends on platform that running tests. Windows use `\`, *nix use `/`, Some other magic platform might use `-` for example. IMHO comparing static string with OS results is not a good idea, So I think that it is good place to make refactoring. But for now I only correct line to avoid failing on main platforms (*nix & windows). Editing all unit-tests is not a deal of this PR. I only make editions for tests don't failure. I think that you can do proper changes after that PR.

src/DacpacTool/PackageBuilder.cs
McHarny: To `if..else` statement we can switch to `switch` statement, it will reduce if'ing on top level.

ErikEJ:
Add timeout to RegEx ...

McHarny:
@jmezach will be any progress?

jmezach:
It looks like CI is failing again and I think I've figured out why. We're using some source code from https://github.com/microsoft/sqltoolsservice and the commit we're using uses a NuGet feed that is no longer available. I think we'll need to update the submodule to use the latest commit on `main` from that repo. @McHarny Can you make that change so that CI will succeed?

McHarny:
@jmezach ok. I will try

McHarny:
@jmezach I see to the `sqltoolsservice` project. Invalid NuGet feed is written in nuget.config, that presents in the latest version of this submodule too. the wlast change of this file is dated by 2019 year.
I think that It need to edit CI definition to not to build this submodule.

jeffrosenberg (changes requested):

src/DacpacTool/Extensions.cs
```diff
@@ -107,26 +107,92 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
...
+            {
+                if(match.Groups["dbl"].Success)
+                    databaseVariableLiteralValue = match.Groups["dbl"].Value;
+                else if (match.Groups["dbv"].Success)
+                    databaseVariableName = match.Groups["dbv"].Value;
+                else if (match.Groups["srv"].Success)
+                    serverVariableName = match.Groups["srv"].Value;
+            }
+
+            if (string.IsNullOrEmpty(serverVariableName) && string.IsNullOrEmpty(databaseVariableName) &&
```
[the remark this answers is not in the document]
McHarny: @jeffrosenberg  Codebase of this is fully taken from `Microsoft.Data.Tools.Schema.Sql.SchemaModel.SqlSchemaCustomSchemaDataWrapper` constructor. If you want's me to refactor this method I can do that. But in generally there are four working combination of these three variables, so four `if`'s will be always present. IMHO

test/DacpacTool.Tests/ScriptParserTests.cs
McHarny: I have Windows 10 Pro 19042.928 installed on my machine. With default Russian locale and Russian language. Also I have language pack installed to .NET SDK. I don't know this is the reason of error or not but the result of test run on picture at attach
Debug

src/DacpacTool/Extensions.cs
```diff
@@ -107,26 +108,90 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
...
+            }
+            else
+            {
+                databaseVariableLiteralValue = Identifier.EncodeIdentifier(externalParts);
+            }
+
+            // if there are no value for database return. Reference MUST have database value
+            if (string.IsNullOrEmpty(databaseVariableLiteralValue) && string.IsNullOrEmpty(databaseVariableName))
+                return null;
+
```
ErikEJ, line 156: nit: Remove empty line

src/DacpacTool/PackageBuilder.cs
```diff
@@ -112,19 +112,16 @@ public bool ValidateModel()
                 if (modelError.Severity == ModelErrorSeverity.Error)
                 {
                     validationErrors++;
+                    Console.WriteLine(modelError.GetOutputMessage(modelError.Severity));
```
ErikEJ, line 115: Why is this part of this PR?

test/DacpacTool.Tests/TestModelBuilder.cs
```diff
@@ -41,9 +41,15 @@ public TestModelBuilder AddStoredProcedureFromFile(string filename)
             return this;
         }
 
-        public TestModelBuilder AddReference(string path)
+        public TestModelBuilder AddReference(string path, string externalParts = "")
```
ErikEJ, line 44: "" => string.Empty

ErikEJ (changes requested):

test/DacpacTool.Tests/TestModelBuilder.cs
McHarny: `string.Empty` doesn't allow to use in parameters default.

src/DacpacTool/PackageBuilder.cs
McHarny: Because of I and my team actively use this great project in our development. Using NuGet packages is what really amazing feature. But we need some features witch presence in SSDT projects and are absent in SqlProj. So I decide to add those features to SqlProj. But approving of them is too slow. All of my additions was developed in two days. As @jmezach ask to split them to partials PR. I do It. but every approve take more than 2 weeks to proceed..... So I think, that It will be more productive to us to maintain our fork separately from the main project. Sorry, I may be newbie for git and GitHub, I've using tfvc for many years, and vss before it. And I began to learn git about some months ago. So switching between branches, rebasing, pull requesting and so, and so, is too hard for me (for now).

ErikEJ (approved):

McHarny:
@jmezach , @jeffrosenberg can you approve?

jeffrosenberg (approved):

jmezach (approved):
LGTM

jmezach:
@McHarny Version 1.14.0 is up on NuGet now which includes these changes.

Issue #27, opened 2020-06-27
Is it possible to add support for GenerateCreateScript?


stevebelskie:
It seems that something in the vein of the example below could be added to `PackageBuilder` or in `Extensions`, that could be called based on a new property added to `PackageOptions`.
```cs
public void GenerateCreateScript(FileInfo dapacFile)
{
  using (var package = DacPackage.Load(dap.FullName))
  using(var file = File.Create("some-parameterized-name.sql"))
  {
     DacServices.GenerateCreateScript(file, package, "test");
               
  }
}
```
Not super familiar with the underlying apis here so there may be a better way using than `DacServices` or a reason that it's not feasible for this project. 

In any case, thanks for this awesome project!

jmezach:
Hi @stevebelskie. Thanks for the suggestion. Would you mind elaborating a little bit about what you are trying to achieve? So far this project has focussed on being able to produce a `.dacpac` from a set of SQL files, while leaving the deployment to existing tools in the ecosystem (namely `SqlPackage.exe`). I wouldn't mind adding an option to generate a `.sql` script as well, but I'm struggling to find a real use case for it.

stevebelskie:
So it's an option available in a `.sqlproj` that is distinct from the option to generate an script when publishing/deploying. The latter I believe is what `SqlPackage.exe` provides. I couldn't figure out a away to generate a script using `SqlPackage.exe` without specifying an actual target db. The `GenerateCreateScript` option (as far as I know) simply creates a sql script that creates the entire database. So as far as I understand it, it creates the `.dapac` in the form of a sql script, it doesn't generate an actual deployment script for some target. 

I'm mostly exploring the idea of using `.dapacs` in furture projects, so this is only somewhat thoughout. The use case for me would be for integration tests run in a docker container. I generally make use of `sqlcmd` to setup an in container db and seed data. It would be nice to be able to run the sql script in those cases so as to avoid having to install `SqlPackage.exe` (the `mssql/server:2019-latest` image comes with `sqlcmd` already) or worry about `.dapac` deployment at all given that I'm just creating and destroying the database in one go.

jmezach:
That makes perfect sense actually. So would you like to have both the script and the .dacpac? I'm not entirely familiar with how this works with the .sqlproj format since I haven't used that feature. Could you briefly explain how it works there, or point me to some doc that explains how it works? Ideally I would like to keep things the same as much as I can.

[...]

PR #133, opened 2021-05-19
Add a new option "--GenerateCreateScript"

markalroberts:
Add a new option "--GenerateCreateScript" which, if specified will generate a complete create script for the dacpac, in a similar way to the way that the SQL Project "create script" tickbox works.

This is useful as it is a very fast operation without doing any compares. It allows integration tests to pick it up and quickly create temporary databases without the overhead of comparing with an empty database, saving a considerable amout of time for large schemas.


markalroberts:
Fixes #27

src/DacpacTool/PackageBuilder.cs
```diff
@@ -388,5 +388,13 @@ private List<int> ParseSuppressionList(string suppressionList)
 
             return result;
         }
+
+        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName)
+        {
+            databaseName ??= Path.GetFileNameWithoutExtension(dacpacFile.Name);
+            using var package = DacPackage.Load(dacpacFile.FullName);
+            using var file = File.Create($"{databaseName}_Create.sql");
+            DacServices.GenerateCreateScript(file, package, databaseName);
```
jeffrosenberg, line 397: Seems simple enough! Would you please add some tests that validate this in action?

src/DacpacTool/Program.cs
```diff
@@ -29,6 +29,7 @@ static async Task<int> Main(string[] args)
                 new Option<string[]>(new string[] { "--sqlcmdvar", "-sc" }, "SqlCmdVariable(s) to include"),
 
                 new Option<bool>(new string[] { "--warnaserror" }, "Treat T-SQL Warnings As Errors"),
+                new Option<bool>(new string[] { "--generatecreatescript" }, "Generate create script for package"),
```
jeffrosenberg, line 32: nit: We generally provide a short form of these arguments as well

jeffrosenberg (changes requested):
Thanks for tackling this! Will you please add some tests to validate the behavior?

jmezach:
I think we'll want to add something to `Sdk.targets` as well so that this behaviour can be triggered from the project file as well so that users don't have to resort to kicking off the tool manually. Not sure what the name of the property is in SQL projects though that controls this.

markalroberts:
> I think we'll want to add something to `Sdk.targets` as well so that this behaviour can be triggered from the project file as well so that users don't have to resort to kicking off the tool manually. Not sure what the name of the property is in SQL projects though that controls this.

Not sure how/if I can (nor whether it's even possible via standard .SqlProj files) - I do not believe the behaviour to generate scripts without targets is exposed via the SQLPackage.exe command line.

jmezach:
@markalroberts When you check the tickbox in a `.sqlproj` I'm sure that adds something to the `.sqlproj` file itself so that the script is generated on build. Or at least that is was I expect to happen.

ErikEJ:
I think that feature is only exposed in the publish dialog, maybe it can be persisted as a publish profile, but not in the .sqlproj

ErikEJ:
I think that feature is only exposed in the publish dialog, maybe it can be persisted as a publish profile, but not in the .sqlproj

jeffrosenberg:
> I do not believe though that there are additional targets I can call, rather that this property is consumed by the build target in the same way as happens when we add the property to this SDK version.

One way or another, I think it's critical that people be able to activate this behavior via the project file, as @jmezach mentioned. It would be nice for it to work in the exact same way as the .sqlproj version, but that's not always possible. If I had to choose between this mirroring the .sqlproj perfectly or being able to activate this behavior via the project file, I would definitely choose the latter.

markalroberts:
> > I do not believe though that there are additional targets I can call, rather that this property is consumed by the build target in the same way as happens when we add the property to this SDK version.
> 
> One way or another, I think it's critical that people be able to activate this behavior via the project file, as @jmezach mentioned. It would be nice for it to work in the exact same way as the .sqlproj version, but that's not always possible. If I had to choose between this mirroring the .sqlproj perfectly or being able to activate this behavior via the project file, I would definitely choose the latter.

I believe this is working the same way as the sqlproj. i.e. there is an option to get it to generate a complete create script as part of the build. I think this is exactly what the issue raised (and this PR) does. No more, no less.

markalroberts:
OK ... so spent a bit of time confirming my beliefs.

There tasks from the `SqlTasks.targets` file used by `SqlProj` are:

```
  <UsingTask TaskName="SqlBuildTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
  <UsingTask TaskName="SqlDeployTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
  <UsingTask TaskName="SqlModelResolutionTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
  <UsingTask TaskName="SqlPublishTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
  <UsingTask TaskName="SqlScriptDependenciesTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
  <UsingTask TaskName="SqlStaticCodeAnalysisTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
  <UsingTask TaskName="SqlSchemaCompareTask" AssemblyFile="$(SqlServerRedistPath)\Microsoft.Data.Tools.Schema.Tasks.Sql.dll" />
```

And the only task that has to option to generate a SQL script is the `SqlBuildTask` task.

Then, examining the `Microsoft.Data.Tools.Schema.Tasks.Sql.dll` file in dotpeek, I can see that the last step on the steps that get run when build is executed:

```
      yield return new Func<bool>(sqlBuildTask.ExecuteBuildIntermediatePaths);
      yield return new Func<bool>(sqlBuildTask.ExecuteClearMessagesStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteProjectPropertiesStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteLoadDatabaseModelPropertiesStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteLoadTaskHostStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteParseSqlCmdVariablesStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteValidateModelStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteBuildContributorsStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteAddSqlCmdVariablesStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteSerializeModelStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteGenerateDeploymentProjectStep);
      yield return new Func<bool>(sqlBuildTask.ExecutePrePostScriptsValidationStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteRefactorLogValidationStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteGeneratePackageStep);
      yield return new Func<bool>(sqlBuildTask.ExecuteGenerateScriptStep);
```

Checks to see if the property is set and generates the script:

```
    private bool ExecuteGenerateScriptStep()
    {
      if (this.GenerateCreateScript)
      {
        ErrorManager errors = new ErrorManager();
        try
        {
          this.CreateScript(this.GetOutputFilepath(this.CreateScriptFileName), errors);
        }
        catch (Exception ex)
        {
          this.ReportError(ex.Message, ErrorManager.ValidationAtBuildCategory);
        }
        this.ReportErrors(errors.GetAllErrors(), ErrorManager.BuildCategory);
      }
      return true;
    }
```

Hopefully this is enough to show what we're trying to achieve with this PR?

jmezach:
@markalroberts I think the changes to the `DacpacTool` are fine and indeed do what is intended here. But in other for someone to actually use it you'd have to invoke the `DacpacTool` manually which I would like to avoid. We should make some changes to `Sdk.targets` to ensure that when the `GenerateCreateScript` property is set to `True` in the project file, that we pass the `--generatecreatescript` option to the `DacpacTool` when we invoke it [here[(https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/1386e339463ab8657da7dc7712d0993070d1f93d/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L222). Then when someone wants to have a script generated at build time they can simply add `<GenerateCreateScript>True</GenerateCreateScript>` to their project file and it will just work.

markalroberts:
OK - thanks, and apologies that I didn't really understand how MSBuild SDKs work.

I think I do better now.

1. I link through the `GenerateCreateScript` property into a command switch for the command line tool
2. I ensure that I create the SQL script in exactly the same path as the dacpac is created (updated tests)
3. I have <Copy> command to copy all SQL artefacts generated alongside the dacpac into the target folder

I have tested this and it works fine.

The only thing I would say is that a clean build does not delete the artefact SQL files, only the compiled output (dacpac) - I personally think this is OK but you may disagree (and perhaps can point me at what I should do!)

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -228,6 +229,12 @@
 
     <!-- Copy any referenced .dacpac packages to the output folder -->
     <Copy SourceFiles="@(IncludedDacpacReferenceFiles)" DestinationFolder="$(TargetDir)" />
+
+    <!-- Copy and SQL artefact created from the command line tool -->
+    <CreateItem Include="$(IntermediateOutputPath)*.sql">
```
jmezach, line 234: I think we can just do the following here:

```xml
<ItemGroup>
  <SqlArtifacts Include="$(IntermediateOutputPath)*.sql" />
</ItemGroup>
```

jmezach:
@markalroberts Don't worry about it, we are all learning :). I'm sorry if my comments came across as too harsh, that wasn't my intent. I truly value your contribution here, just trying to make it the best as possible.

As for your question, yes, I think we'll want to clean up the SQL files as well as part of the `Clean` target. We can probably fix that by adding a `<Target Name="CleanSqlArtifacts" AfterTargets="AfterClean">` and use the [DeleteTask](https://docs.microsoft.com/en-gb/previous-versions/visualstudio/visual-studio-2015/msbuild/delete-task?view=vs-2015&redirectedfrom=MSDN) to remove the additional files.

jmezach (approved):
This looks good to me. Just one final thing, perhaps we should add the new property to the [test project](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/test/TestProject/TestProject.csproj). That will make sure this feature is run in the CI pipeline as well.

test/TestProject/TestProject.csproj
```diff
@@ -9,6 +9,7 @@
     <ReadCommittedSnapshot>True</ReadCommittedSnapshot>
     <ServiceBrokerOption>EnableBroker</ServiceBrokerOption>
     <PackageProjectUrl>https://github.com/jmezach/MSBuild.Sdk.SqlProj/</PackageProjectUrl>
+	<GenerateCreateScript>True</GenerateCreateScript>
```
jmezach, line 12: Indentation is a little bit off ;).

markalroberts: Aaah tabs and spaced :) will fix that (and change notepad++ to do spaces!)

src/DacpacTool/PackageBuilder.cs
```diff
@@ -388,5 +388,13 @@ private List<int> ParseSuppressionList(string suppressionList)
 
             return result;
         }
+
+        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName)
```
jmezach, line 392: Hmm, I was just thinking we might want to write something to the console here so that we can see that this is actually happening?

jmezach:
@jeffrosenberg Does this look good to you now?

jeffrosenberg (approved):
LGTM, thanks again for picking this up!

markalroberts:
Thanks for taking a bit of time to help me make the right changes, and for making this very useful library!

Will this get published to nuget as a new version any time soon? Would be very useful for us to be able to make use of this...

ErikEJ (approved):

jmezach:
@markalroberts Just merged this PR. I'll put out a release soon. Thank you for your contribution!

ErikEJ:
@jmezach Maybe update to latest DacFX first?

jmezach:
;) Too late I'm afraid. Anything important in there?

ErikEJ:
I mean before a new release.

I think it is wise to always use the latest release version, but it is closed source, and there are newer any release notes, sadly.

PR #134, opened 2021-05-27
Fix bug with SQL create script not copying to output

markalroberts:
It turns out that when run via MSBuild (not in VS) the item group SqlArtifacts is not automatically updated with the SQL artifacts from the build, so the sql script does not copy into the target folder (just sits in obj subfolder)


src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -235,11 +231,17 @@
...
     <!-- Copy any SQL artifacts created from the command line tool -->
+    <CreateItem Include="$(IntermediateOutputPath)*.sql">
+        <Output TaskParameter="Include" ItemName="SqlArtifacts" />
+    </CreateItem>
     <Copy SourceFiles="@(SqlArtifacts)" DestinationFolder="$(TargetDir)" />
   </Target>
 
   <!--Ensure that any SQL artifacts generated by the compile are cleaned up properly -->
   <Target Name="CleanSqlArtifacts" AfterTargets="AfterClean">
+      <CreateItem Include="$(IntermediateOutputPath)*.sql">
```
jeffrosenberg, line 242: Is this block meant to appear here? It's a duplicate of the block above, and we have both a `CreateItem` and a `Delete`, so I just wanted to confirm.

test/TestProjectWithScript/TestProjectWithScript.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <Import Project="$(MSBuildThisFileDirectory)..\..\src\MSBuild.Sdk.SqlProj\Sdk\Sdk.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <SqlServerVersion>Sql150</SqlServerVersion>
+    <RecoveryMode>Simple</RecoveryMode>
+    <AllowSnapshotIsolation>True</AllowSnapshotIsolation>
+    <ReadCommittedSnapshot>True</ReadCommittedSnapshot>
+    <ServiceBrokerOption>EnableBroker</ServiceBrokerOption>
+    <PackageProjectUrl>https://github.com/jmezach/MSBuild.Sdk.SqlProj/</PackageProjectUrl>
```
jeffrosenberg, line 11: nit: should be github.com/rr-wfm/

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -32,10 +32,6 @@
     <IntermediateRefAssembly Remove="@(IntermediateRefAssembly)" />
   </ItemGroup>
 
-  <ItemGroup>
```
jeffrosenberg, line 35: Just to make sure I understand, this PR is meant to fix issues when running on MSBuild. If you remove this block, does it work for both MSBuild and VS? Or does changing this break VS?

markalroberts: Hi @jeffrosenberg - the problem I am seeing is that the creation script is generated fine but only actually copied to output in VS - when run on TeamCity (using MSBuild), for some reason the sql file is not progressing beyond the `obj` folder to the target output folder.

I *think* that this is because the "SqlArtifacts" `itemgroup` is not updated after the sql is generated by the CoreCompile... it seems that if we use CreateItem then it does pick up the new file.

Could be that @jmezach will have some thoughts on this as I think it was he that suggested I use the ItemGroup ...

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
markalroberts: Well, I guess I think that if we need the use the CreateItem to pick up the SQL to copy, then we may also need to use it to pick up SQL to delete.

I suppose another way is to keep the itemgroup as-is but just _update_ it after the create script creates the sql

markalroberts:
I've double-checked this and can confirm that `<itemgroup>` is evaluated on script startup so we need to modify/create an itemgroup to pick up artifacts after the build.

I've run this in MSBuild and can see it working properly now.

I've also checked msbuild /t:clean to confirm that the files are deleted from obj folder _and_ target

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
markalroberts: I tried various combinations but the createitem seems to be the most reliable when it comes to (a) picking up any new artifacts created by build and (b) being run in isolation (e.g. msbuild /t:clean)

markalroberts:
Sorry to be adding this in after the original PR to generate scripts - I have only just been able to properly see the problem as it was masked by Visual Studio locally.

test/TestProjectWithScript/TestProject.nuspec
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
jmezach, line 1: I don't think we really need the `.nuspec` here.

test/TestProject/TestProject.csproj
```diff
@@ -9,7 +9,6 @@
     <ReadCommittedSnapshot>True</ReadCommittedSnapshot>
     <ServiceBrokerOption>EnableBroker</ServiceBrokerOption>
     <PackageProjectUrl>https://github.com/jmezach/MSBuild.Sdk.SqlProj/</PackageProjectUrl>
-    <GenerateCreateScript>True</GenerateCreateScript>
```
jmezach, line 12: Was there a specific reason for moving this out into a separate project? Since we're not building any of the other projects as part of CI currently this feature is not being tested end-to-end in CI (except for the unit tests)

test/TestProject/TestProject.csproj
```diff
@@ -9,7 +9,6 @@
     <ReadCommittedSnapshot>True</ReadCommittedSnapshot>
     <ServiceBrokerOption>EnableBroker</ServiceBrokerOption>
     <PackageProjectUrl>https://github.com/jmezach/MSBuild.Sdk.SqlProj/</PackageProjectUrl>
```
jmezach, line 11: `jmezach` should probably be replaced with `rr-wfm` here while we're at it ;).

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jmezach: It seems that MSBuild is evaluating the `ItemGroup` once and then doesn't update it again. I guess in that scenario `CreateItem` makes sense.

jmezach (approved):
LGTM. Just a couple of comments.

test/TestProject/TestProject.csproj
markalroberts: @jmezach  - yes.... do now that the artifact is copied to properly, any projects that reference this project (e.g. `TestProjectWithSDKRef`) will bring in not only the .dacpac but also the .sql file - the code in the DacpacTool will then see the .sql file as part of a file to incorporate into a new dacpac (which is wrong) and DacFx produces errors when trying to add this file.

So, I made a new project to test project to exercise just this functionality without breaking anything else as nothing will reference the new project.

jmezach: That might actually be an issue. Perhaps we should ignore any `.sql` files in the `bin` and `obj` folders by default as they could have been generated by the build so we'll probably don't want to include them.

jeffrosenberg: Yes, I would definitely agree with this. In my projects at work, we always add `Content Remove` nodes for `bin/` and `obj/`. Not sure why it didn't occur to me to suggest this as a broader change to the framework -- I can't see why we would ever want to include these.

jeffrosenberg (approved):

test/TestProject/TestProject.csproj
markalroberts: I have created an issue for this - happy to make a separate PR for it?

https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/137

jmezach: Makes sense. And yes, a PR for this is more than welcome :).

markalroberts:
Is it worth getting this merged and released or do people think there is more to do here?

jmezach:
No, this looks good to me. I'll get this released.

jmezach:
@markalroberts Version 1.16.0 is up on NuGet now. Took a little bit more time due to some GitHub issues unfortunately. I've also incorporated some dependency updates, hence the minor version rather than a bug fix. If necessary we can always port the fix to 1.15 of course, but if this works for you I don't see much reason to do so.

markalroberts:
Thank you for doing this.... I can confirm that it now works in our rather convoluted build process, which is a good sign - we have fast localdb setups and integration tests all in .net 5 which is brilliant! Thanks.

PR #166, opened 2021-07-08
Improve error message.

### LINKED ISSUE #165: SDK stops working if NET6.0 (preview) is installed
As the title says, this project does not work if NET6.0 SDK is installed, even if side-by-side with NET5.0 (which is supported) or earlier supported frameworks. This is due to the checks here:

https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/4d872dd45edcca5a741ae6855d81276ff25df3ad/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L92-L98

I was able to side-step the issue by adding a global.json file to my repo that capped the SDK to the 5.0.x line, however I'm wondering if there's something we can do here. I'm not asking for 6.0 support since it's still in preview, but I'm wondering if there's some way for this SDK to force itself to a *supported, installed* framework. I don't know the particulars of the MSBuild/SDK system so this may not be possible.... But it was quite confusing when my build started failing simply because I installed the preview SDK on my machine (side-by-side with 5.0).

A related issue that I noticed due to all of this is the error message on line 98 which says this sdk only supports 2.1/3.1:
https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/4d872dd45edcca5a741ae6855d81276ff25df3ad/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L98-L98

This SDK does support NET5.0...it seems like the error message was missed during that update.

ErikEJ:
fixes #165

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -95,7 +95,7 @@
       <DacpacToolExe>$(MSBuildThisFileDirectory)../tools/$(DacpacToolTfm)/DacpacTool.dll</DacpacToolExe>
     </PropertyGroup>
     <Error Condition="!Exists($(DacpacToolExe))"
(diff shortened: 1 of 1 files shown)


jmezach (approved):

taskadmin (approved):

jmezach:
@jeffrosenberg Could you have a look at this as well?

jeffrosenberg (approved):
Approved. Apologies for the delay, I was out of town without internet access 🙂

jmezach:
No problem. Thanks for the review.

Issue #117, opened 2021-01-26
Adding package reference to DACPAC project does not result in .dacpac file being copied to bin directory


flakey-bit:
In my solution (simplified) I have two projects:
- `Some.Namespace.Database` - this is the `MSBuild.Sdk.SqlProj` SDK project for producing the DACPAC
- `Some.Namespace.DataAccess` - this is a vanilla `Microsoft.NET.Sdk` (`netstandard2.0`) project containing C# data access code

I've added a project reference from `Some.Namespace.DataAccess` to  `Some.Namespace.Database` as follows:

```
  <ItemGroup>
    <ProjectReference Include="..\Some.Namespace.Database\Some.Namespace.Database.csproj" />
  </ItemGroup>
```

When I build `Some.Namespace.Database` it produces the `.dacpac` file in the output directory for that project, however when I build `Some.Namespace.DataAccess` I was hoping/expecting the `.dacpac` file to be copied to the output directory for _that_ project, since it's a dependency.

Is there any easy way to achieve what I want? NB use-case is around programatically loading the DACPAC file via `DacServices` / `DacPackage.Load` in `Microsoft.SqlServer.Dac` (NuGet package [Microsoft.SqlServer.DacFx.x64](https://www.nuget.org/packages/Microsoft.SqlServer.DacFx.x64/))

NB: I tried changing `Some.Namespace.DataAccess` to a concrete framework (`netcoreapp3.1`) however that didn't help 😢

jmezach:
@flakey-bit This isn't currently build in unfortunately. In fact I'm even surprised that it builds at all because I would expect your project `Some.Namespace.DataAccess` to not compile since it would try to include a reference to the `.dacpac` file which obviously doesn't work.

That being said I do think this could be a reasonable scenario to support. According to the docs for the [project reference protocol](https://github.com/dotnet/msbuild/blob/master/documentation/ProjectReference-Protocol.md) we might be able to implement a `GetCopyToOutputDirectoryItems` target that would return the `.dacpac` which should then copy that file to the output directory.

If you're adventurous we would appreciate a PR for this. Not sure when we'll have time to implement this.

ErikEJ:
A post build step could maybe also work?

AroglDarthu:
This one was interesting... @jmezach is correct with his assumption that the project does not compile. However this can be achieved by adding `<ReferenceOutputAssembly>False</ReferenceOutputAssembly>`:
```
[...]

PR #189, opened 2021-09-11
Allow project reference from vanilla classlib

AroglDarthu:
This solves #117 

Take care to add `<ReferenceOutputAssembly>False</ReferenceOutputAssembly>` to the project reference. Without it, the compilation of the vanilla class library will fail; a `.dacpac` file is not a valid assembly ;-)

The target `GetCopyToOutputDirectoryItems` returns an `AllItemsFullPathWithTargetPath` item collection. The new target `CopyDacpacs` runs before `GetCopyToOutputDirectoryItems` and preloads `AllItemsFullPathWithTargetPath` with the project's dacpac file and all referenced dacpac files.

Two test projects prove that the referenced dacpac files get copied to the bin folder of the vanilla class library project:
- TestIncludeFromVanillaProj
This uses a `ProjectReference` to a dacpac project with a nested `ProjectReference`.
- TestIncludeFromVanillaProjWithExternalPackageReference
This uses a 'ProjectReference' to a dacpac project with a nested dacpac `PackageReference`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,6 +471,17 @@ DECLARE @Query NVARCHAR(MAX) = '<your-script>'
 EXEC (@Query)
 ```
 
+## Reference `MSBuild.Sdk.SqlProj` from class library
+The output of `MSBuild.Sdk.SqlProj` is not an assembly, but a `.dacpac`. In order to correctly reference a `MSBuild.Sdk.SqlProj` based project from a class library, the `ReferenceOutputAssembly` hint needs to be set to `False`:
+```
+<ItemGroup>
+    <ProjectReference
+      Include="../MyDacpacProj/MyDacpacProj.csproj"
+      ReferenceOutputAssembly="False" />
+</ItemGroup>
+```
+Now, upon compilation of the class library, the relevant `.dacpac` files get copied to the output directory.
+
 ## Known limitations
 Since this is not an entire project system but only an MSBuild SDK we cannot provide IntelliSense for objects defined within the project. This limitation can be circumvented by connecting the SQL editor to a live database that is used for development purposes.
 
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -279,6 +279,7 @@
     <Exec Command="$(DacpacToolCommand)" />
   </Target>
   
+  <!-- Lists all dacpac files of interest to the current project, so that a referencing class library can copy these to its output directory -->
   <Target Name="CopyDacpacs" BeforeTargets="GetCopyToOutputDirectoryItems">
     <ItemGroup>
       <CopyDacpacFiles Include="$(TargetPath)"/>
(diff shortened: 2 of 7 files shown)


src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -279,4 +279,17 @@
     <Exec Command="$(DacpacToolCommand)" />
   </Target>
   
+  <Target Name="CopyDacpacs" BeforeTargets="GetCopyToOutputDirectoryItems">
```
jmezach, line 282: We should probably add some comments here explaining what this target is for for future reference.

test/TestIncludeFromVanillaProj/TestIncludeFromVanillaProj.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+      <ProjectReference Include="../TestProjectWithWarnings/TestProjectWithWarnings.csproj">
+        <ReferenceOutputAssembly>False</ReferenceOutputAssembly>
```
jmezach, line 9: I believe we can have the `ReferenceOutputAssembly` as an attribute on the `ProjectReference` element as well which is a bit cleaner IMHO.

test/TestIncludeFromVanillaProj/TestIncludeFromVanillaProj.csproj
```diff
@@ -0,0 +1,17 @@
...
+      </ProjectReference>
+
+      <ProjectReference Include="../TestProjectWithProjectReference/TestProjectWithProjectReference.csproj">
+        <ReferenceOutputAssembly>False</ReferenceOutputAssembly>
```
jmezach, line 13: We should probably explain this in the README as well.

jmezach (changes requested):

jmezach (approved):
LGTM. Perhaps @ErikEJ and @jeffrosenberg could have a look at this as well.

README.md
```diff
@@ -471,6 +471,17 @@ DECLARE @Query NVARCHAR(MAX) = '<your-script>'
 EXEC (@Query)
 ```
 
+## Reference `MSBuild.Sdk.SqlProj` from class library
+The output of `MSBuild.Sdk.SqlProj` is not an assembly, but a `.dacpac`. In order to correctly reference a `MSBuild.Sdk.SqlProj` based project from a class library, the `ReferenceOutputAssembly` hint needs to be set to `False`:
+```
+<ItemGroup>
+    <ProjectReference
+      Include="../MyDacpacProj/MyDacpacProj.csproj"
```
ErikEJ, line 479: Should it be ".sqlproj" ??

ErikEJ (approved):

jmezach (approved):

jmezach:
Let's get this merged!

Issue #174, opened 2021-08-09
Embed README in NuGet package


jmezach:
Given the new [design](https://www.nuget.org/packages/MSBuild.Sdk.SqlProj/1.16.2?preview=1) on NuGet.org it might be a good idea to embed the README in the NuGet package itself. It takes a more prominent approach in the new design, so it could help people get started quicker when they find the package on NuGet.org.

PR #190, opened 2021-09-12
Include README.md in the nuget package

AroglDarthu:
Implemented as per https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/
@jmezach Can you please [verify](https://docs.microsoft.com/en-us/nuget/nuget-org/publish-a-package#web-portal-use-the-upload-package-tab-on-nugetorg) that the readme shows up?

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.nuspec
+++ b/src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.nuspec
@@ -14,10 +14,12 @@
       <packageType name="$packagetype$" />
     </packageTypes>
     <repository type="git" url="https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/" />
+    <readme>docs\README.md</readme>
   </metadata>
   <files>
     <file src="Sdk\**" target="Sdk" />
     <file src="tools\**" target="tools" />
     <file src="LICENSE.txt" target="LICENSE.txt" />
+    <file src="..\..\README.md" target="docs\" />
   </files>
 </package>


ErikEJ (approved):

src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.nuspec
```diff
@@ -14,10 +14,12 @@
       <packageType name="$packagetype$" />
     </packageTypes>
     <repository type="git" url="https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/" />
+    <readme>docs\README.md</readme>
   </metadata>
   <files>
     <file src="Sdk\**" target="Sdk" />
     <file src="tools\**" target="tools" />
     <file src="LICENSE.txt" target="LICENSE.txt" />
+    <file src="..\..\README.md" target="docs\" />
```
ErikEJ, line 23: Should it not just be: ..\README.md ?

ErikEJ (changes requested):

src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.nuspec
AroglDarthu: Mmm... I was just wondering where the `LICENCE.txt` comes from. A target (SetNuSpecProperties) in the project file copies it over from the solution root. I could do the same trick for the readme.
Just wondering if there is an added benefit?

AroglDarthu: The solution root is two levels up from the project file, so `..\..\README.md` is correct. However, I'ld like to keep it in line with what is already there. So I'll copy it to the nuspec location first, just as the `LICENSE` file..

AroglDarthu: On the other hand... this might lead to inadvertently modifying the wrong README.md.
Maybe it's better for the README.md to not copy it?

ErikEJ: Just leave it as is, I think

ErikEJ (approved):

src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.nuspec
AroglDarthu: Sorry, which version do you mean? The first, so `..\..\README.md` or the second one with the copy action?

ErikEJ: Original

jmezach: If I remember correctly the reason the LICENSE.txt is copied first is because the original file is called just LICENSE which NuGet cannot deal with (a file without an extension) so that's why we copy it as LICENSE.txt so that it can be packaged.

jmezach:
Fixes #174

jmezach:
I've just uploaded the prerelease version from this PR to NuGet.org and the README.md shows up as expected, so this seems to work just fine. I'll go ahead and merge this.

jmezach (approved):

Issue #192, opened 2021-09-14
Unknown target framework moniker


AroglDarthu:
https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/e57fc2470b80ac277d2ccaf6b1336580ec3c88c9/src/DacpacTool/DacpacTool.csproj#L5

`netcoreapp5.0` should be `net5.0`
The test project targets net5.0. In Visual Studio that leads to the error message:
> Project '..\..\src\DacpacTool\DacpacTool.csproj' targets 'netcoreapp2.1;netcoreapp3.1;netcoreapp5.0'. It cannot be referenced by a project that targets '.NETFramework,Version=v5.0'.

jmezach:
Hmm, I believe that `netcoreapp5.0` and `net5.0` are actually synonymous. The reason we're using `netcoreapp5.0` is because of [this check](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L92).

I haven't had any problems with it so far to be honest, although I haven't opened the solution in Visual Studio at all. I usually use VS Code which doesn't seem to have any issues with it.

AroglDarthu:
You could be right in that `netcoreapp5.0` _can_ still be used. The error that I got, made me look for it in the docs. The moniker `netcoreapp5.0` is no longer official: https://docs.microsoft.com/en-us/dotnet/standard/frameworks

On the bright side: I got the error because of a newer .NET5 SDK and an old(er) version of Visual Studio.

The following line will also need to be changed:
https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/e57fc2470b80ac277d2ccaf6b1336580ec3c88c9/src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj#L34

And then of course the line you mentioned (and the error message below it) should use `MicrosoftNETBuildTasksTFM` which translates to `netcoreapp2.1` or `netcoreapp3.1` or `net5.0`.

jmezach:
[...]

PR #194, opened 2021-09-16
Remove support for netcoreapp2.1 and add support for net6.0

AroglDarthu:
Solves #192 and adds support for net6.0

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,12 +26,7 @@ jobs:
         fetch-depth: 0 # avoid shallow clone so nbgv can do its work.
         submodules: 'true' # need to check out the SqlTools submodule to build successfully.
 
-    # Install .NET Core SDK's for 2.1, 3.1 and 5.0
-    - name: Setup dotnet 2.1
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: '2.1.x'
-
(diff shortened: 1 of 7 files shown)


AroglDarthu:
Help wanted with following problem: https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/192#issuecomment-921158637

AroglDarthu:
@jmezach The build pipeline does not yet support `net6.0 sdk`

jmezach:
@AroglDarthu We can fix the pipeline by modifying [the workflow](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/.github/workflows/main.yml). We should probably remove .NET Core 2.1 there as well and add .NET 6. Be aware that there are a couple of places where we install .NET SDK's in there.

Really appreciate you helping us out with this!

AroglDarthu:
Cool. Never looked at github workflows before. I'll do some reading on it in the weekend ;-)

jmezach:
Looks like this is finally working, so I think it will be a good idea to merge to master and ship a preview release for folks that are on .NET 6 already while we wait for a stable version of DACFX before we ship a stable version. @ErikEJ @jeffrosenberg Perhaps you guys can give this a review. Thanks to @AroglDarthu for his initial work on this.

ErikEJ (approved):

jeffrosenberg (approved):
LGTM

PR #224, opened 2021-11-26
Include composite objects from external references when generating script file

dovic95:
Add an `IncludeCompositeObjects` option that allows to include objects from the referenced database packages when generating the script file

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,7 +453,8 @@ Instead of using `dotnet publish` to deploy changes to a database, you can also 
 ```xml
 <Project Sdk="MSBuild.Sdk.SqlProj/1.15.0">
   <PropertyGroup>
-    <GenerateCreateScript>True</GenerateCreateScript>
+      <GenerateCreateScript>True</GenerateCreateScript>
+      <IncludeCompositeObjects>True</IncludeCompositeObjects>
   </PropertyGroup>
 </Project>
 ```
@@ -464,7 +465,7 @@ The database name for the create script gets resolved in the following manner:
 1. Package name.
 > Note: 
 >- the generated script also uses the resolved database name via a setvar command.
->- the composite objects (tables, etc.) from external references are also included in the generated script   
+>- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script   
 
 ## Workaround for parser errors (SQL46010)
 This project relies on the publicly available T-SQL parser which may not support all T-SQL syntax constructions. Therefore you might encounter a SQL46010 error if you have a script file that contains unsupported syntax. If that happens, there's a couple of workarounds you can try:
--- a/README.md
+++ b/README.md
@@ -462,7 +462,9 @@ With this enabled you'll find a SQL script with the name `<database-name>_Create
 The database name for the create script gets resolved in the following manner:
 1. `TargetDatabaseName`.
 1. Package name.
-> Note: the generated script also uses the resolved database name via a setvar command.
+> Note: 
+>- the generated script also uses the resolved database name via a setvar command.
+>- the composite objects (tables, etc.) from external references are also included in the generated script   
 
 ## Workaround for parser errors (SQL46010)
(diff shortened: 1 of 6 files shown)


src/DacpacTool/PackageBuilder.cs
```diff
@@ -401,7 +401,9 @@ public void GenerateCreateScript(FileInfo dacpacFile, string databaseName)
 
             using var package = DacPackage.Load(dacpacFile.FullName);
             using var file = File.Create(Path.Combine(dacpacFile.DirectoryName, scriptFileName));
-            DacServices.GenerateCreateScript(file, package, databaseName);
+
+            var options = new DacDeployOptions() { IncludeCompositeObjects = true };
```
jmezach, line 405: It looks like we're hardcoding this to be true now which might not be what everyone wants. Ideally we'd pass the value of the MSBuild property down to here and use that.

jmezach (changes requested):
@dovic95 Thank you for your contribution. I just have one comment.

src/DacpacTool/PackageBuilder.cs
dovic95: Yes, you're right. I changed this,

README.md
```diff
@@ -462,7 +463,9 @@ With this enabled you'll find a SQL script with the name `<database-name>_Create
 The database name for the create script gets resolved in the following manner:
 1. `TargetDatabaseName`.
 1. Package name.
-> Note: the generated script also uses the resolved database name via a setvar command.
+> Note: 
+>- the generated script also uses the resolved database name via a setvar command.
+>- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script
```
jmezach, line 468: Perhaps it is worth noting here that `IncludeCompositeObjects` also defaults to true (see [here](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props#L217)).

src/DacpacTool/PackageBuilder.cs
```diff
@@ -389,7 +389,7 @@ private List<int> ParseSuppressionList(string suppressionList)
             return result;
         }
 
-        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName)
+        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName, bool includeCompositeObjects = false)
```
jmezach, line 392: Is there a reason for defining a default here? This is only called from one location right?

jmezach (approved):
Just a couple of small comments

src/DacpacTool/PackageBuilder.cs
dovic95: Removed as it is called from a unique place. The other calls are from unit tests

README.md
dovic95: Done

jmezach:
LGTM. @ErikEJ @jeffrosenberg Perhaps either one of you can take a look here as well?

src/DacpacTool/PackageBuilder.cs
```diff
@@ -161,7 +161,7 @@ public void SaveToDisk(FileInfo outputFile, PackageOptions packageOptions = null
             EnsureModelCreated();
             EnsureModelValidated();
             EnsureMetadataCreated();
-
```
jeffrosenberg, line 164: Looks like there are a bunch of unnecessary whitespace changes. Not the end of the world, but it would be nice to fix these.

jeffrosenberg (approved):
I'd prefer that you fix the whitespace issues before merging, if you can do that relatively easily, since it confuses the commitment history and a future `git blame`. Code looks good, though, so I'm going to just make that a suggestion and approve.

src/DacpacTool/PackageBuilder.cs
dovic95: Fixed. This was due to my Rider settings

Issue #230, opened 2021-12-04
GenerateCreateScript: Set CreateNewDatabase to true


dovic95:
Hello,

First, @jmezach and @jeffrosenberg thank you for accepting my pull request #224 

I tested the new version in my project and it now produces a generate script that includes tables from referenced `dacpac`s. The only 'issue' is the `CreateNewDatabase` property not set, the script does not create the database when I run it (I have to manually create the database first..). A bit annoying

I will propose a new pull request where the `CreateNewDatabase` can be set. The question is: Should we provide the possibility to set it through a property (name of this property?) or should it be always `true`?

jmezach:
I'm starting to feel that for the `GenerateCreateScript` functionality we might want to pass along all the `    KnownDeployProperties` (see [here](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/5d12a407eeb7ca734d690417cc6a8ba2c64387f0/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props#L122)) because they could possibly all influence the outcome of the generated script and folks might want to set them.

dovic95:
That makes sense. 

Should we provide the DeployOptions properties object to the [BuildDacpac](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/5d12a407eeb7ca734d690417cc6a8ba2c64387f0/src/DacpacTool/Program.cs#L76) method (so the different `KnownDeployProperties `will be provided) or should we create a specific `CommandHandler `for generating the script (that will run after `BuildDacpac`) ?

jmezach:
We'll probably want to do it as part of building as that would probably be the fastest.

rizi:
> We'll probably want to do it as part of building as that would probably be the fastest.

Does this mean a db is never created automatically? Or it won't be created when it comes from a referenced assembly?

I would like that a db should be created when it does not already exist, does this work at the moment? (we don't have references at the moment).

Br

jmezach:
@rizi I believe that by default a new database is created if one doesn't exist. `CreateNewDatabase=True` basically means that if a database exists, it is dropped first and a new one is created.

rizi:
> @rizi I believe that by default a new database is created if one doesn't exist. `CreateNewDatabase=True` basically means that if a database exists, it is dropped first and a new one is created.

Thank you very much for fast response!

Br

dovic95:
[...]

PR #244, opened 2021-12-27
Pass down KnownDeployProperties to the GenerateCreateScript functionality

dovic95:
This PR allows setting all the possible properties in the `GenerateCreateScript` functionality. The properties defined in the `csproj` (the `KnownDeployProperties` ones) are now automatically set when generating the create script

Solves issue #230

### DIFF
--- a/src/DacpacTool/BuildOptions.cs
+++ b/src/DacpacTool/BuildOptions.cs
@@ -11,8 +11,8 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public SqlServerVersion SqlServerVersion { get; set; }
         public FileInfo InputFile { get; set; }
         public string[] Reference { get; set; }
-        public string[] Property { get; set; }
-        public string[] CreateScriptProperty { get; set; }
+        public string[] BuildProperty { get; set; }
+        public string[] DeployProperty { get; set; }
         public string[] SqlCmdVar { get; set; }
         public FileInfo PreDeploy { get; set; }
         public FileInfo PostDeploy { get; set; }
--- a/src/DacpacTool/BuildOptions.cs
+++ b/src/DacpacTool/BuildOptions.cs
@@ -12,6 +12,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public FileInfo InputFile { get; set; }
         public string[] Reference { get; set; }
         public string[] Property { get; set; }
+        public string[] CreateScriptProperty { get; set; }
         public string[] SqlCmdVar { get; set; }
         public FileInfo PreDeploy { get; set; }
         public FileInfo PostDeploy { get; set; }
@@ -21,7 +22,6 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public string SuppressWarnings { get; set; }
         public FileInfo SuppressWarningsListFile { get; set; }
         public bool GenerateCreateScript { get; set; }
-        public bool IncludeCompositeObjects { get; set; }
         public string TargetDatabaseName { get; set; }
     }
 }
--- a/src/DacpacTool/BuildOptions.cs
+++ b/src/DacpacTool/BuildOptions.cs
@@ -21,6 +21,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public string SuppressWarnings { get; set; }
         public FileInfo SuppressWarningsListFile { get; set; }
         public bool GenerateCreateScript { get; set; }
+        public bool IncludeCompositeObjects { get; set; }
         public string TargetDatabaseName { get; set; }
     }
 }
--- a/src/DacpacTool/PackageBuilder.cs
+++ b/src/DacpacTool/PackageBuilder.cs
@@ -389,7 +389,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             return result;
         }
 
-        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName)
+        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName, bool includeCompositeObjects = false)
         {
             if (string.IsNullOrWhiteSpace(databaseName))
             {
@@ -402,7 +402,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             using var package = DacPackage.Load(dacpacFile.FullName);
             using var file = File.Create(Path.Combine(dacpacFile.DirectoryName, scriptFileName));
 
-            var options = new DacDeployOptions() { IncludeCompositeObjects = true };
+            var options = new DacDeployOptions() { IncludeCompositeObjects = includeCompositeObjects };
             DacServices.GenerateCreateScript(file, package, databaseName, options);
         }
     }
--- a/src/DacpacTool/PackageBuilder.cs
+++ b/src/DacpacTool/PackageBuilder.cs
@@ -161,7 +161,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             EnsureModelCreated();
             EnsureModelValidated();
             EnsureMetadataCreated();
-
+            
             // Check if the file already exists
             if (outputFile.Exists)
             {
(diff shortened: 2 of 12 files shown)


src/DacpacTool/PackageBuilder.cs
```diff
@@ -389,21 +389,21 @@ private List<int> ParseSuppressionList(string suppressionList)
             return result;
         }
 
-        public void GenerateCreateScript(FileInfo dacpacFile, string databaseName, bool includeCompositeObjects)
+        public void GenerateCreateScript(BuildOptions options)
```
jmezach, line 392: I'd like to keep the seperation of concerns here and not pass in the `BuildOptions` directly, but keep the seperate parameters. I guess we could just replace the boolean with DacDeployOptions.

src/DacpacTool/PackageDeployer.cs
```diff
@@ -213,12 +213,12 @@ public void SetProperty(string key, string value)
                     "CompareUsingTargetCollation" => bool.Parse(value),
                     "CreateNewDatabase" => bool.Parse(value),
                     "DatabaseLockTimeout" => int.Parse(value),
-                    "DatabaseSpecification" => ParseDatabaseSpecification(value),
+                    "DatabaseSpecification" => PropertyParser.ParseDatabaseSpecification(value),
```
jmezach, line 216: Can't we replace this method with the `PropertyParser` to avoid duplication of logic?

src/DacpacTool/BuildOptions.cs
```diff
@@ -12,6 +12,7 @@ public class BuildOptions : BaseOptions
         public FileInfo InputFile { get; set; }
         public string[] Reference { get; set; }
         public string[] Property { get; set; }
+        public string[] CreateScriptProperty { get; set; }
```
jmezach, line 15: I'm a bit on the fence about the name of this property. Technically they are used when generating the create script, but essentially they are deployment options, so perhaps `DeployProperty` would be better name? And then rename `Property` to `BuildProperty` as well, to make the distinction clear?

test/DacpacTool.Tests/PackageBuilderTests.cs
```diff
@@ -605,17 +610,69 @@ public void GenerateCreateScript_IncludeCompositeObjects(bool includeCompositeOb
...
+
+            // Cleanup
+            tempFile.Delete();
+        }
+
+        [TestMethod]
+        [DataRow(true)]
+        [DataRow(false)]
+        [DataRow(null)]
+        public void GenerateCreateScript_CreateNewDatabase(bool? createNewDatabase)
```
jmezach, line 644: It feels a bit weird to have these seperate test cases for what is essentially the same logic, which is making sure we can pass along deployment properties.

test/DacpacTool.Tests/PropertyParserTest.cs
```diff
@@ -0,0 +1,176 @@
+﻿// <copyright file="PropertyParserTest.cs" company="Suez">
```
jmezach, line 1: Please remove this copyright notice.

src/DacpacTool/PropertyParser.cs
```diff
@@ -0,0 +1,160 @@
+﻿
```
jmezach, line 1: Please remove this useless newline.

jmezach (changes requested):
@dovic95 Thank you for your contribution. I have a couple of comments.

src/DacpacTool/BuildOptions.cs
dovic95: I've renamed the properties as suggested

src/DacpacTool/PackageBuilder.cs
dovic95: Done. I agree it's better to separate the two concepts and apply the Hollywood principle :)

src/DacpacTool/PackageDeployer.cs
dovic95: Done

test/DacpacTool.Tests/PropertyParserTest.cs
dovic95: Removed

test/DacpacTool.Tests/PackageBuilderTests.cs
dovic95: I can remove these tests but I think we should be 100% sure the deployment options are not modified inside the `GenerateCreateScript` method.

I can add a test where I first generate the script using `DacServices.GenerateCreateScript` method and then compare it to the one generated by the `GenerateCreateScript` method.

src/DacpacTool/PropertyParser.cs
```diff
@@ -0,0 +1,149 @@
...
+            CustomParsers.Add("ExcludeObjectTypes", ParseObjectTypes);
+            CustomParsers.Add("DatabaseSpecification", ParseDatabaseSpecification);
+        }
+
+        /// <summary>
+        /// Converts the <see cref="BuildOptions"/> to a <see cref="DacDeployOptions"/> object with all properties
+        /// </summary>
+        /// <param name="options">The build options</param>
+        /// <returns>The <see cref="DacDeployOptions"/> object</returns>
+        public static DacDeployOptions ToDacDeployOptions(this BuildOptions options)
```
jmezach, line 30: `ToDacDeployOptions` suggests we are converting one into the other, while actually we're just extracting a subset of the `BuildOptions` instance into a `DacDeployOptions` object. Perhaps `ExtractDeployOptions` would be a more appropriate name?

test/DacpacTool.Tests/PackageBuilderTests.cs
jmezach: Yeah, I think that makes sense. My comment was more about that we now have 2 separate tests that trigger the same logic, but the only difference is setting a different property on the `DacDeployOptions` class.

src/DacpacTool/PackageDeployer.cs
```diff
@@ -192,118 +190,37 @@ private void HandleDacServicesMessage(object sender, DacMessageEventArgs args)
...
         }
 
-        public void SetProperty(string key, string value)
+        public void SetDeployProperty(string deployProperty)
+            => this.SetDeployProperties(new[] { deployProperty });
+
+        public void SetDeployProperties(string[] deployProperties)
         {
-            try
+            foreach (var deployProperty in deployProperties)
```
jmezach, line 198: Is it just me, or does it still feel like we're duplicating some logic here that's also in the `PropertyParser` class now?

test/DacpacTool.Tests/PackageBuilderTests.cs
dovic95: I've removed the two tests and added one that checks the expected create script

jmezach: Looks good :)

src/DacpacTool/PackageDeployer.cs
dovic95: Where do you see the duplication? The `foreach` ?

dovic95: Mmmh we could actually have only one method for setting properties in a `DacDeployOptions` object. I just have a doubt regarding this verification and if it's applicable during the build:

```
                if (databaseProperty.Name == "SqlCommandVariableValues")
                {
                    throw new ArgumentException("SQLCMD variables should be set using the --sqlcmdvar command line argument and not as a property.");
                }
```

I'll have a look in a few hours

dovic95: I've mutualized the `SetDeployProperties` method

src/DacpacTool/PropertyParser.cs
dovic95: I've renamed the method.

dovic95:
@jmezach I've resolved all the comments. Let me know if the PR is OK for you

src/DacpacTool/PropertyParser.cs
jmezach: It seems this newline is still here?

dovic95: I don't see it on my side:
image

jmezach: Nevermind, I wasn't looking straight ;). Must be the end of the year.

jmezach:
@dovic95 Looks good to me.

@ErikEJ @jeffrosenberg Could either of you have a look at this as well, thanks!

ErikEJ:
LGTM

jmezach (approved):

dovic95:
Hi, are we waiting for another approval? Would be great if we can complete this PR and release the version :)

jeffrosenberg (approved):
LGTM

PR #284, opened 2022-04-21
Dev/add condition for nuspec

### LINKED ISSUE #283: Automatic Nuspec generation
https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/d80ce593feb255471cf4b60e46a520e67b52bedb/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props#L29

The reason a `.nuspec` file needs to be created manually to get `dotnet pack` to work is the fact you are setting the `<NuspecFile>$(MSBuildProjectName).nuspec</NuspecFile>` property.  If you either remove this line or add a condition to check for the existence of the file `dotnet pack` will correctly generate the `.nuspec` for you.  

A simple work around until this project is updated is to just clear out `<NuspecFile></NuspecFile>` in one of your project file's property groups.

mwwhited:
Fixes #283

These changes allow for generating nuspec instead of having to provide one.

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
@@ -26,7 +26,7 @@
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
     <NoPackageAnalysis>True</NoPackageAnalysis>
-    <NuspecFile>$(MSBuildProjectName).nuspec</NuspecFile>
+    <NuspecFile Condition="Exists('$(MSBuildProjectName).nuspec')">$(MSBuildProjectName).nuspec"</NuspecFile>
   </PropertyGroup>
 
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" Condition=" '$(MicrosoftCommonPropsHasBeenImported)' != 'true' "/>
@@ -35,6 +35,19 @@
 
   <!-- For CPS/VS support. Importing in .props allows any subsequent targets to redefine this if needed -->
   <Target Name="CompileDesignTime" />
+    
+    <Target Name="SFCustomizePacking" AfterTargets="CoreBuild">
+        <ItemGroup Label="DacpacFilesToCopyBuild" Condition="'$(IncludeAllDacPacs)'!='true'">
+            <Content Include="$(OutputPath)\$(MSBuildProjectName).dacpac" Pack="true" PackagePath="tools/">
+                <PackageCopyToOutput>true</PackageCopyToOutput>
+            </Content>
+        </ItemGroup>
+        <ItemGroup Label="DacpacFilesToCopyAll" Condition="'$(IncludeAllDacPacs)'=='true'">
+            <Content Include="$(OutputPath)\*.dacpac" Pack="true" PackagePath="tools/">
+                <PackageCopyToOutput>true</PackageCopyToOutput>
+            </Content>
+        </ItemGroup>
+    </Target>
 
   <ItemGroup>
     <Content Include="**/*.sql">
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
@@ -37,16 +37,19 @@
   <Target Name="CompileDesignTime" />
     
     <Target Name="SFCustomizePacking" AfterTargets="CoreBuild">
-        <ItemGroup Label="DacpacFilesToCopyBuild" Condition="'$(IncludeAllDacPacs)'!='true'">
-            <Content Include="$(OutputPath)$(MSBuildProjectName).dacpac" Pack="true" PackagePath="tools/">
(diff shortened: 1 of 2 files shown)


src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -52,16 +52,17 @@
...
+          description=$(PackageDescription);
+          version=$(PackageVersion);
+          authors=$(Authors);
+          copyright=$(Copyright);
+          projecturl=$(PackageProjectUrl);
+          tags=$(PackageTags);
+          configuration=$(Configuration);
+          tfm=$(TargetFramework);
+          packagetype=Dacpac;
+          $(NuspecProperties)
```
mwwhited, line 65: This change allows for additional properties to be passed in when using the nuspec file.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
```diff
@@ -26,7 +26,7 @@
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
     <NoPackageAnalysis>True</NoPackageAnalysis>
-    <NuspecFile>$(MSBuildProjectName).nuspec</NuspecFile>
+    <NuspecFile Condition="Exists('$(MSBuildProjectName).nuspec')">$(MSBuildProjectName).nuspec"</NuspecFile>
```
mwwhited: This removes the requirement for the nuspec file to exist allowing the msbuild sdk to generate the nuspec file

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
```diff
@@ -35,6 +35,19 @@
 
   <!-- For CPS/VS support. Importing in .props allows any subsequent targets to redefine this if needed -->
   <Target Name="CompileDesignTime" />
+    
+    <Target Name="SFCustomizePacking" AfterTargets="CoreBuild">
```
mwwhited, line 39: this will add the single project dacpac or all dacpacs in the build folder to be added to the nuget package automatically.

mwwhited:
Converted back to draft as there looks to be an issue with how I'm building the file paths when executed on azure devops.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
thibault-reigner: I face the same issue (custom `<NuspecProperties>` from csproj file being ignored) and reached the same conclusion. I can confirm this fixes the issue.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -46,22 +46,24 @@
       IncrementalClean;
       PostBuildEvent
     </CoreBuildDependsOn>
-    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);SetNuSpecProperties</GenerateNuspecDependsOn>
+    <GenerateNuspecDependsOn>$(GenerateNuspecDependsOn);SetNuSpecProperties;_PackageDatabaseResults</GenerateNuspecDependsOn>
+    <PackageType Condition="'$(PackageType)'==''" >dacpac</PackageType>
```
jmezach, line 50: Hmm, I seem to remember that actually setting the package type breaks some things as you can no longer install these packages anymore. Regardless though we're not actually using this value currently.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
```diff
@@ -35,6 +35,35 @@
 
   <!-- For CPS/VS support. Importing in .props allows any subsequent targets to redefine this if needed -->
   <Target Name="CompileDesignTime" />
+    
+	<Target Name="_PackageDatabaseResults" Condition="'$(IsDatabaseProject)'=='True'" BeforeTargets="_GetPackageFiles">
```
jmezach, line 39: What does this section do exactly?

mwwhited: This control the build dependency order.  When I had multiple projects without this feature they were sometime building out of order and not collecting all the child dependencies when the build was running multithreaded.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
mwwhited: Without doing this some times the packages were being built incorrectly and didn't not have all of their children referenced correctly.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
mikeBulava: Hate to be a Pain, but any idea when this will be ready to merge??  Does it work?  Currently I need to pack these projects differently than standard projects until this feature is completed...

mwwhited: It worked when I sent the pr a few months back.  I have a work around for my own projects so if you guys don't want to accept the change and own it then no worries from me.

PR #286, opened 2022-04-30
Allow sql projects to publish on build (PublishOnBuild).

AraHaan:
This is triggered with a new msbuild property: PublishOnBuild.

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -258,6 +258,9 @@
       <Delete Files="@(SqlArtifacts -> '$(TargetDir)%(Filename)%(Extension)')" />
   </Target>
 
+  <!-- Allows a project to set to publish on build for their sql project (so they can deploy it on build that is if they want). -->
+  <Target Name="PublishSqlProject" BeforeTargets="CoreCompile" DependsOnTargets="Publish" Condition="'$(PublishOnBuild)' != 'False' And '$(DeployOnPublish)' != 'False'" />
+
   <!-- Performs a publish of the .dacpac produced by build to a configurable target server -->
   <Target Name="Publish" DependsOnTargets="Build" Condition="'$(DeployOnPublish)' != 'False'">
     <ItemGroup>
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -259,7 +259,7 @@
   </Target>
 
   <!-- Allows a project to set to publish on build for their sql project (so they can deploy it on build that is if they want). -->
-  <Target Name="PublishSqlProject" BeforeTargets="CoreCompile" DependsOnTargets="Publish" Condition="'$(PublishOnBuild)' != 'False' And '$(DeployOnPublish)' != 'False'" />
+  <Target Name="PublishSqlProject" BeforeTargets="CoreCompile" DependsOnTargets="Publish" Condition="'$(PublishOnBuild)' != 'False'" />
 
   <!-- Performs a publish of the .dacpac produced by build to a configurable target server -->
   <Target Name="Publish" DependsOnTargets="Build" Condition="'$(DeployOnPublish)' != 'False'">


src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -302,4 +305,4 @@
     </ItemGroup>
     <Message Importance="Low" Text="CopyDacpacFiles: @(CopyDacpacFiles)" />
   </Target>
-</Project>
```
AraHaan, line 305: I blame GitHub for this line.

AraHaan:
cc: @jmezach

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -258,6 +258,9 @@
       <Delete Files="@(SqlArtifacts -> '$(TargetDir)%(Filename)%(Extension)')" />
   </Target>
 
+  <!-- Allows a project to set to publish on build for their sql project (so they can deploy it on build that is if they want). -->
+  <Target Name="PublishSqlProject" BeforeTargets="CoreCompile" DependsOnTargets="Publish" Condition="'$(PublishOnBuild)' != 'False' And '$(DeployOnPublish)' != 'False'" />
```
jmezach, line 262: Looking at the `Condition` here I get the feeling that it is an opt-out now, rather than an opt-in. To be honest I think this should be an opt-in.

Additionally it seems we have two separate properties that achieve the same thing?

AraHaan: I could do it like that however it seems the Publish target runs when ``DeployOnPublish`` is ``true``, while this might try to run the target it might not run if the user sets ``DeployOnPublish`` to ``false``.

AraHaan: Removed ``DeployOnPublish`` from this line. I wonder if it should also be removed from the publish target and instead checked within it where it will try to deploy it if it's not set.

ErikEJ:
How is this different from

    dotnet build

    dotnet publish

especially since we support publish for local development only

AraHaan:
It allows this to be published even with the sqlproj just happens to be part of a solution where publishing those projects *might* fail (because they depend on runtimes for which the only supported form of publish is FDD).

ErikEJ:
I do not understand this sentence: _because they depend on runtimes for which the only supported form of publish is FDD_ - what is FDD? 

>> Also, it acts as a shortcut for when you only want to run build on the entire solution and do not want to run publish on the entire solution as well (however you want to publish specific projects in it via cli).

How is that very different from just

    dotnet build

    dotnet publish

AraHaan:
> I do not understand this sentence: _because they depend on runtimes for which the only supported form of publish is FDD_ - what is FDD?
> 
> > > Also, it acts as a shortcut for when you only want to run build on the entire solution and do not want to run publish on the entire solution as well (however you want to publish specific projects in it via cli).
> 
> How is that very different from just
> 
> ```
> dotnet build
> 
> dotnet publish
> ```

Framework dependent deployment (or FDD). I think ``dotnet publish`` not specifies a RID since the release of the .NET 7 SDK which implies self-contained.

Imran-imtiaz48 (commented):
sadad

Imran-imtiaz48 (commented):
The old and new code snippets appear to serve similar purposes, focusing on publishing a SQL project during the build process based on certain conditions (PublishOnBuild and DeployOnPublish variables). Here’s a breakdown of the changes:

Simplification: The new code has removed the condition And '$(DeployOnPublish)' != 'False' from the PublishSqlProject target, simplifying the logic. This change suggests a streamlined approach, potentially reducing complexity and making the code easier to understand.

Functionality Retention: Both versions retain functionality related to publishing the .dacpac file to a target server after the build (Publish target).

Clarity: The removal of the additional condition in the PublishSqlProject target in the new code improves clarity by focusing directly on the PublishOnBuild condition, which may enhance maintainability and readability of the build script.

Conclusion:

The new code improves upon the old code by simplifying the condition in the PublishSqlProject target while retaining the essential functionality of publishing the .dacpac file. This change not only enhances clarity but also potentially reduces the risk of unintended logic errors, making the build script more robust and easier to manage.

ErikEJ:
@jmezach Closing this due to inactivity - and proper justification IMHO - feel free to re-open if you think otherwise

PR #363, opened 2022-11-22
Use DacFx for SQL Server 2022

ErikEJ:
Use microsoft.sqlpackage on container

### DIFF
--- a/src/DacpacTool/PackageDeployer.cs
+++ b/src/DacpacTool/PackageDeployer.cs
@@ -18,7 +18,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             _console = console ?? throw new ArgumentNullException(nameof(console));
         }
 
-        public SqlConnectionStringBuilder ConnectionStringBuilder { get; private set; } = new SqlConnectionStringBuilder { Encrypt = false };
+        public SqlConnectionStringBuilder ConnectionStringBuilder { get; private set; } = new SqlConnectionStringBuilder();
         public DacDeployOptions DeployOptions { get; private set; } = new DacDeployOptions();
 
         public void UseTargetServer(string targetServer)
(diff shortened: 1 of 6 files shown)


src/DacpacTool/PackageDeployer.cs
```diff
@@ -18,7 +18,7 @@ public PackageDeployer(IConsole console)
             _console = console ?? throw new ArgumentNullException(nameof(console));
         }
 
-        public SqlConnectionStringBuilder ConnectionStringBuilder { get; private set; } = new SqlConnectionStringBuilder();
+        public SqlConnectionStringBuilder ConnectionStringBuilder { get; private set; } = new SqlConnectionStringBuilder { Encrypt = false };
```
jmezach, line 21: Hmm, do we want to hardcode this? I think we should at least make this configurable.

ErikEJ: Is this not just setting the default value?

ErikEJ: we have not set "trust server certificate" before either. And this makes publish work just like before - for local dev publishing

jmezach: Yeah, I guess it is fine as a default, but then we should at least document it I think. Perhaps there is someone out there who does want to encrypt the connection to his/her development SQL Server and now they don't have a way of doing that. They can make it configurable if they want, but then at least they know it is happening ;).

ErikEJ: How about adding it to DeployOptions and default to false ?

public bool Encrypt { get; set; }

jmezach: Yeah, that will work :).

jmezach (approved):

jeffrosenberg (approved):

ErikEJ:
@jmezach Is it OK for me to merge this now? (or do you need to do something behind the scenes) ?

jmezach:
No, I think we can merge this. After that I can do a release.

Issue #20, opened 2020-05-06
SqlCmdVariables should be in a single section


ErikEJ:
Expected:

                <CustomData Category="SqlCmdVariables" Type="SqlCmdVariable">
			<Metadata Name="DbReaderPassword" Value="" />
			<Metadata Name="DbUserPassword" Value="" />
		</CustomData>

Actual:

                <CustomData Category="SqlCmdVariables" Type="SqlCmdVariable">
			<Metadata Name="DbReaderPassword" Value="" />
		</CustomData>
		<CustomData Category="SqlCmdVariables" Type="SqlCmdVariable">
			<Metadata Name="DbUserPassword" Value="" />
		</CustomData>

(Have not tested actually using the variables, blocked on

jmezach:
Looking at the current implementation this makes sense. Shouldn't be too hard to fix though. We'll probably have to supply a list to `AddSqlCmdVariable` (and rename it to be plural) and then call the `SetMetadata` method multiple times for each variable, before adding the custom data to the model.

PR #368, opened 2022-11-24
Load Default SQLCmd Parameters value into dacpac

### LINKED ISSUE #248: SQLCMD Variable Defaults not Present in Built DacPac
We are using sql command variables to define the server and database on a project reference, but the default values are not making their way into the dacpacs when we do a `dotnet build`. We tried using the `dotnet build /p:SqlCmdVar__1=somevalue` but it didn't net a difference. Using v.1.17.3. 

Our sqlproj:
```
<Project Sdk="MSBuild.Sdk.SqlProj/1.17.3">
    <PropertyGroup>
        <TargetFramework>netstandard2.1</TargetFramework>
        <RecoveryMode>Simple</RecoveryMode>
        <SqlServerVersion>Sql130</SqlServerVersion>
    </PropertyGroup>
    <PropertyGroup>
        <OutputPath>deploy\</OutputPath>
        <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
        <AppendRuntimeIdentifierToOutputPath>false</AppendRuntimeIdentifierToOutputPath>
    </PropertyGroup>
    <ItemGroup>
        <Content Remove="scripts\**\*.sql" />
    </ItemGroup>
    <ItemGroup>
        <ProjectReference Include="../Soft_Landing/Soft_Landing.sqlproj" DatabaseVariableLiteralValue="Soft_Landing" />
        <ProjectReference Include="../Soft_App/Soft_App.sqlproj" DatabaseVariableLiteralValue="Soft_App" />
        <ProjectReference Include="../ACK/ACK.sqlproj" DatabaseSqlCmdVariable="ACK" ServerSqlCmdVariable="EXTServer" />
    </ItemGroup>
    <ItemGroup>
        <SqlCmdVariable Include="ACK">
            <DefaultValue>ACK</DefaultValue>
            <Value>$(SqlCmdVar__1)</Value>
        </SqlCmdVariable>
        <SqlCmdVariable Include="EXTServer">
            <DefaultValue>192.168.1.1</DefaultValue>
            <Value>$(SqlCmdVar__2)</Value>
        </SqlCmdVariable>
    </ItemGroup>
</Project>
```
Then when examining the model.xml of the dacpac:
```
...

		<CustomData Category="SqlCmdVariables" Type="SqlCmdVariable">
			<Metadata Name="ACK" Value="" />
			<Metadata Name="EXTServer" Value="" />
		</CustomData>
	</Header>
```

We are using Azure Data Studio to perform the SQL Compare from the built dacpacs but they fail due the values on these vars missing.

I also tried adding the explicit values in the variables at the top of the sqlproj to no avail.
```
<PropertyGroup>
      <SqlCmdVar__1>ACK</SqlCmdVar__1>
</PropertyGroup>
```

Geims83:
fixes #248

### DIFF
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -204,22 +204,13 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             return result;
         }
 
-        public static void AddSqlCmdVariables(this TSqlModel model, string[] variables)
+        public static void AddSqlCmdVariable(this TSqlModel model, string variableName, string variableDefaultValue)
         {
             var dataSchemaModel = GetDataSchemaModel(model);
 
             var customData = Activator.CreateInstance(Type.GetType("Microsoft.Data.Tools.Schema.SchemaModel.CustomSchemaData, Microsoft.Data.Tools.Schema.Sql"), "SqlCmdVariables", "SqlCmdVariable");
-
-            foreach (var variable in variables)
-            {
-                var values = variable.Split('=');
-                var variableName = values[0];
-                string variableDefaultValue = string.Empty;
-                if (values.Length > 1)
-                    variableDefaultValue = values[1];
-                var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
-                setMetadataMethod.Invoke(customData, new object[] { variableName, variableDefaultValue });
-            }
+            var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
+            setMetadataMethod.Invoke(customData, new object[] { variableName, variableDefaultValue });
 
             AddCustomData(dataSchemaModel, customData);
         }
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -212,11 +212,18 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
 
             foreach (var variable in variables)
             {
-                var values = variable.Split('=');
-                var variableName = values[0];
+                var varWithValue = variable.Split('=', 2);
+                var variableName = varWithValue[0];
                 string variableDefaultValue = string.Empty;
-                if (values.Length > 1)
-                    variableDefaultValue = values[1];
+
+                if (varWithValue.Length > 1 && varWithValue[1] != string.Empty)
+                {
+                    variableDefaultValue = varWithValue[1];
+                    Console.WriteLine($"Adding SqlCmd variable {variableName} with default value {variableDefaultValue}");
+                }
+                else
+                    Console.WriteLine($"Adding SqlCmd variable {variableName}");
+
                 var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
                 setMetadataMethod.Invoke(customData, new object[] { variableName, variableDefaultValue });
             }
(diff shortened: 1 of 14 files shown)


ErikEJ:
You should be able to add a test, have a look at AddSqlCmdvariable test

Geims83:
Tests added! Thank you for the hint.

ErikEJ (approved):
LGTM

jeffrosenberg (approved):

src/DacpacTool/PackageBuilder.cs
```diff
@@ -50,17 +50,27 @@ private void ValidateReference(string referenceFile)
...
         }
 
-        public void AddSqlCmdVariables(string[] variableNames)
+        public void AddSqlCmdVariables(string[] variables)
         {
             // Ensure that the model has been created
             EnsureModelCreated();
 
-            foreach (var variableName in variableNames)
+            foreach (var variable in variables)
```
jmezach, line 58: Seems like this `foreach` loop only exists for logging purposes now, so we're essentially enumerating the list of variables twice. Perhaps we should move this logging into the `AddSqlCmdVariables()` method in `Extensions` instead?

jmezach (changes requested):
Made one comment. Also, I was thinking we might want to add default values to the test project that uses SQLCMD variables as well.

src/DacpacTool/PackageBuilder.cs
Geims83: What about keeping this loop and removing the one in `Extensions`, with a new `AddSqlCmdVariable(string Name, string DefaultValue)` method to 'Extensions'? Feels cleaner to me.

jmezach: Yeah, I agree that does feel cleaner.

Geims83:
@jmezach fixed the issue in the comment.

> I was thinking we might want to add default values to the test project that uses SQLCMD variables as well.

It seems to me that there are already default values in sqlcmd variables in the projects where they are used. Am I missing something?

jmezach:
Yes, you're right, sorry about that.

src/DacpacTool/Extensions.cs
```diff
@@ -204,17 +204,13 @@ public static IEnumerable<string> GetReferencedDacPackages(this TSqlModel model)
...
             var dataSchemaModel = GetDataSchemaModel(model);
 
             var customData = Activator.CreateInstance(Type.GetType("Microsoft.Data.Tools.Schema.SchemaModel.CustomSchemaData, Microsoft.Data.Tools.Schema.Sql"), "SqlCmdVariables", "SqlCmdVariable");
-
-            foreach (var variableName in variableNames)
-            {
-                var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
-                setMetadataMethod.Invoke(customData, new object[] { variableName, string.Empty });
-            }
+            var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
```
jmezach, line 212: Now that I look at this I'm wondering if this works the way we expect it to work. Previously we would only add one `CustomSchemaData` object to the model with all the variables. Now we're adding it multiple times for each variable. To be honest I don't know if that matters, but I think we should have a look at the resulting `.dacpac` from both a `.sqlproj` and an `MSBuild.Sdk.SqlProj` project to make sure we're producing the correct output.

Geims83: Just found out issue #20 that was about this problem, so I reverted the commit and moved console logging inside `Extensions` as you were suggesting.

jmezach (approved):
LGTM

PR #397, opened 2023-02-16
Allow overriding the dacpac filename

### LINKED ISSUE #396: Allow overriding dacpac filename for a package reference
When adding a `PackageReference` to a project, MSBuild.Sdk.SqlProj currently assumes that this package contains a `.dacpac` in the `tools` folder which has the same name as the package. So for example, if I add a `<PackageReference Include="MyPackage">` to the project file, it will look for a file called `tools\MyPackage.dacpac` and use that as a reference. If such a file does not exist, the package reference is essentially ignored.

While that is a fair assumption that works for most scenarios, and is in fact also the result of doing a `dotnet pack` on an MSBuild.Sdk.SqlProj project, there are some scenarios where this doesn't hold up. One example is the newly released [Microsoft.SqlServer.Dacpacs](https://www.nuget.org/packages/Microsoft.SqlServer.Dacpacs) package, which holds the `.dacpac` for the `master` and `msdb` databases. Ideally you should be able to reference those in an `MSBuild.Sdk.SqlProj` project, but you can't do that currently.

I would therefore like to introduce the possibility to override the name of the `.dacpac` on a per `PackageReference` basis. For example, something like this:

```xml
<PackageReference Include="Microsoft.SqlServer.Dacpacs" DacpacPath="tools\master.dacpac" />
```

jmezach:
Microsoft has released NuGet packages containing the `.dacpac` of the `master` and `msdb` databases. They decided to ship a single `Microsoft.SqlServer.Dacpacs` package containing both of them however, which `MSBuild.Sdk.SqlProj` currently has no way of supporting as we're assuming that the name of the `.dacpac` is the same as package.

To remedy this I've added support for overriding the name of the `.dacpac` within the package on a per-package reference basis. I believe this is a generally useful feature, but works especially well with these newly released packages.

Fixes #396

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ dotnet new sqlproj -s Sql130
 You should now have a project file with the following contents:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.2.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
         <SqlServerVersion>Sql130</SqlServerVersion>
@@ -92,7 +92,7 @@ If you already have a SSDT (.sqlproj) project in your solution, you can keep tha
 There are a lot of properties that can be set on the model in the resulting `.dacpac` file which can be influenced by setting those properties in the project file using the same name. For example, the snippet below sets the `RecoveryMode` property to `Simple`:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.2.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
         <RecoveryMode>Simple</RecoveryMode>
@@ -110,7 +110,7 @@ Like `.sqlproj` projects  `MSBuild.Sdk.SqlProj` supports controlling T-SQL build
 Treating warnings as errors can be optionally enabled by adding a property `TreatTSqlWarningsAsErrors` to the project file:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.2.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
     <PropertyGroup>
         <TreatTSqlWarningsAsErrors>True</TreatTSqlWarningsAsErrors>
         ...
@@ -122,7 +122,7 @@ Treating warnings as errors can be optionally enabled by adding a property `Trea
 To suppress specific warnings from being treated as errors, add a comma-separated list of warning codes to `SuppressTSqlWarnings` property in the project file:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.2.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
     <PropertyGroup>
         <SuppressTSqlWarnings>71558,71502</SuppressTSqlWarnings>
         <TreatTSqlWarningsAsErrors>True</TreatTSqlWarningsAsErrors>
@@ -134,7 +134,7 @@ To suppress specific warnings from being treated as errors, add a comma-separate
 You can suppress warnings for a specific file by adding `SuppressTSqlWarnings` for this file:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.2.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
     <PropertyGroup>
         ...
     </PropertyGroup>
@@ -154,7 +154,7 @@ Support for pre- and post deployment scripts has been added in version 1.1.0. Th
 To include these scripts into your `.dacpac` add the following to your `.csproj`:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.2.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
     <PropertyGroup>
         ...
     </PropertyGroup>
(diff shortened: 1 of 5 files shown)


ErikEJ (approved):
LGTM

README.md
```diff
@@ -222,10 +222,29 @@ Especially when using pre- and post deployment scripts, but also in other scenar
...
+        <SqlServerVersion>Sql160</SqlServerVersion>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.SqlServer.Dacpacs" Version="160.0.0" DacpacName="master" />
+    </ItemGroup>
+</Project>
+```
+
+This will add a reference to the `tools\master.dacpac` file inside the [Microsoft.SqlServer.Dacpacs](https://www.nuget.org/packages/Microsoft.SqlServer.Dacpacs) package. Note that if that file doesn't exist within the package, the package reference will still be silently ignored. However the build will most likely fail if your project actually references objects from the reference package.
```
jeffrosenberg, line 240: I feel like `Microsoft.SqlServer.Dacpacs` references are important enough for users that maybe they should get their own section. Here you're combining the documentation for the `DacpacName` attribute with documentation for using `Microsoft.SqlServer.Dacpacs`, but I think maybe each deserves its own section. Having a heading for "Microsoft system databases" or something similar would make this a lot easier for users to find. Just my personal opinion, feel free to disagree :slightly_smiling_face:

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -118,7 +118,8 @@
         -->
         <PhysicalLocation Condition="'%(_ResolvedPackageReference.PhysicalLocation)'==''">$([System.String]::new('$(NuGetPackageRoot)%(PackageReference.Identity)/%(PackageReference.Version)').ToLower())</PhysicalLocation>
 
-        <DacpacFile>%(_ResolvedPackageReference.PhysicalLocation)/tools/%(Identity).dacpac</DacpacFile>
+        <DacpacFile>%(_ResolvedPackageReference.PhysicalLocation)/tools/%(PackageReference.Identity).dacpac</DacpacFile>
+        <DacpacFile Condition="Exists('%(_ResolvedPackageReference.PhysicalLocation)/tools/%(PackageReference.DacpacName).dacpac')">%(_ResolvedPackageReference.PhysicalLocation)/tools/%(PackageReference.DacpacName).dacpac</DacpacFile>
```
jeffrosenberg, line 122: Is there any potential issue with having the `DacpacFile` element appear twice? I know we've said that if the file doesn't exist, it just gets silently ignored. What if a user did something like `<PackageReference Include="MyPackage" Version="160.0.0" DacpacName="MyPackage" />`? In this case, we'd have two equivalent `DacpacFile` elements, would that break anything?

test/TestProjectWithMasterReference/Procedures/csp_MyProcedureThatUsesMaster.sql
```diff
@@ -0,0 +1,6 @@
+CREATE PROCEDURE [dbo].[csp_MyProcedureThatUsesMaster]
+    @p_SomeParam int
```
jeffrosenberg, line 2: nit: this parameter isn't needed. Doesn't hurt anything either, so no big deal, just thought I'd point it out!

jeffrosenberg (approved):
Looks good! Neat that this can be done directly through the `.targets` file. I did have a few comments, but they're all pretty minor, so I'll approve.

README.md
jmezach: Yeah, I was a bit on the fence about that, but I think it makes sense to put it into its own section.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jmezach: I think in that case it will just overwrite it, so it shouldn't be a problem. That being said it is probably a good idea to test it ;). I'll make sure we have that covered in one of the test projects somewhere.

test/TestProjectWithMasterReference/Procedures/csp_MyProcedureThatUsesMaster.sql
jmezach: Sure, I've removed it :).

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jmezach: I've modified the `TestProjectWithPackageReference` test project to have a `PackageReference` to `TestProject` with `DacpacName=TestProject` and that works like a charm.

README.md
jmezach: Fixed it now.

README.md
```diff
@@ -279,11 +296,31 @@ sqlpackage
...
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.SqlServer.Dacpacs" Version="160.0.0" DacpacName="master" />
+    </ItemGroup>
+</Project>
+```
+
+The above example references the `master` database from the [Microsoft.SqlServer.Dacpacs](https://www.nuget.org/packages/Microsoft.SqlServer.Dacpacs) NuGet package. Please note that there are different versions of that package for different versions of SQL Server. It is recommended to reference the same version of the package as the `SqlServerVersion` you are targeting with your project, as seen in the example above.
+
+For the Azure SQL and Synapse variants of SQL Server there are dedicated packages: [Microsoft.SqlServer.Dacpacs.Azure](https://www.nuget.org/packages/Microsoft.SqlServer.Dacpacs.Azure) and [Microsoft.SqlServer.Dacpacs.Synapse]https://www.nuget.org/packages/Microsoft.SqlServer.Dacpacs.Synapse) respectively.
```
ErikEJ, line 317: missing (

README.md
```diff
@@ -279,11 +296,31 @@ sqlpackage
     /Properties:IncludeCompositeObjects=True
 ```
 
+## Referencing system databases
+Microsoft has recently released NuGet packages containing the definitions of the `mater` and `msdb` databases. This is useful if you want to reference objects from those databases within your own projects without getting warnings. To reference these, you'll need to use at least version 2.5.0 of MSBuild.Sdk.SqlProj as you'll need to use the `DacpacName` feature for package references described above. For example:
```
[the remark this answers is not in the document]
jmezach: Thanks :). Fixed it.

jmezach:
I think we can get this merged?

ErikEJ:
Yes!

PR #466, opened 2023-09-04
Update docs re system .dacpacs

ErikEJ:
Update readme to latest release version

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ dotnet new sqlproj -s Sql130
 You should now have a project file with the following contents:
 
 ```xml
-<Project Sdk="MSBuild.Sdk.SqlProj/2.5.0">
+<Project Sdk="MSBuild.Sdk.SqlProj/2.6.1">
     <PropertyGroup>
         <TargetFramework>netstandard2.0</TargetFramework>
         <SqlServerVersion>Sql130</SqlServerVersion>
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -224,10 +224,10 @@ Especially when using pre- and post deployment scripts, but also in other scenar
 </Project>
 ```
 
-It will assume that the `.dacpac` file is inside the `tools` folder of the referenced package and that it has the same name as the NuGet package. Referenced packages that do not adhere to this convention will be silently ignored. However, you have the ability to override this convention by using the `DacpacName` attribute on the `PackageReference` (introduced in version 2.5.0). For example:
+It will assume that the `.dacpac` file is inside the `tools` folder of the referenced package and that it has the same name as the NuGet package. Referenced packages that do not adhere to this convention will be silently ignored. However, you have the ability to override this convention by using the `DacpacName` attribute on the `PackageReference` (introduced in version 2.6.1). For example:
```
jeffrosenberg, line 227: This instance of `2.5.0` probably shouldn't be changed, right?

README.md
```diff
@@ -299,30 +299,30 @@ sqlpackage
 ```
 
 ## Referencing system databases
-Microsoft has recently released NuGet packages containing the definitions of the `master` and `msdb` databases. This is useful if you want to reference objects from those databases within your own projects without getting warnings. To reference these, you'll need to use at least version 2.5.0 of MSBuild.Sdk.SqlProj as you'll need to use the `DacpacName` feature for package references described above. For example:
+Microsoft has recently released NuGet packages containing the definitions of the `master` and `msdb` databases. This is useful if you want to reference objects from those databases within your own projects without getting warnings. To reference these, you'll need to use at least version 2.6.1 of MSBuild.Sdk.SqlProj as you'll need to use the `DacpacName` feature for package references described above. For example:
```
jeffrosenberg, line 302: Same here, leave at 2.5.0

jeffrosenberg (approved):
Approved, but I noted 2 spots that would be good to fix before merging

ErikEJ:
@jeffrosenberg Thanks, updated

jmezach (approved):
LGTM

Issue #170, opened 2021-07-27
Question: Compatibility with SSDT rules?


tcartwright:
I wrote some rules that we use during our build process for SSDT. I was wondering if you or anyone else have tested out if the rules still work with your projects. I am very interested in being able to use nuget with SSDT.

The rules I wrote are here, if you are interested: https://github.com/tcartwright/SqlServer.Rules

Sorry for using the issues to submit you a question.

jmezach:
Hi @tcartwright. First of all thank you for asking this question. I don't have a lot of experience with custom rules so I haven't tried this myself yet. That being said I can't really think of a reason why it can't work, it is more a matter of figuring out how to make it work. Looking at your README it seems that just placing the assemblies in the right location would trigger them at build them for regular SQL Database Projects. Is that correct? Or do you need to put something into the project file in any way?

tcartwright:
Hey @jmezach, thank you very much for the response. I appreciate it very much. Getting the rules in place is actually a two step process. Really should update the readme. :) Once you place the assemblies in place you turn on custom code analysis in the project properties. You will also need to select any rules that are unselected in the list or deselect any you don't want to run. If you apply my custom rules in the correct directory beforehand, they should also show up in the list of rules to run. Once code analysis is enabled then just perform a rebuild. Let me know if you want / need any help with it. 

I would love to get your project in use at my current company. Like you I did a lot of work with SSDT. Our stories sound very similar. I build deployment system around TFS build / Octopus Deploy.  I also integrated t-sqlt unit tests into our build with code coverage. I was highly interested when I read your blog post and what you had done. 

NOTE: Once you run a build with code analysis enabled VS will place a hard lock on the rule assemblies. To update them or remove them you will need to shut down any SSDT projects Visual Studios.

jmezach:
@tcartwright Can you share the changes made to the project file after you've enabled your custom rules? That might give some idea for how we might go about enabling this in MSBuild.Sdk.SqlProj.

As for your NOTE, if we're able to get this working I'm guessing that hard lock goes away since the build phase happens outside of Visual Studio with MSBuild.Sdk.SqlProj. Unfortunately that also means that you only get the warnings on build and not while editing your code.

tcartwright:
@jmezach here are the changes made to a normal SSDT project to enable the analysis:

Turned on, all rules enabled:
`    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>`

[...]

PR #515, opened 2024-02-25
Analyzers

ErikEJ:
implements parts of  #170 

(Still need SDK / build support)

Includes support for the built-in rules and the list of rules here: https://github.com/tcartwright/SqlServer.Rules/blob/master/docs/table_of_contents.md

### DIFF
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview1" />
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview2" />
     <PackageReference Include="Microsoft.SqlServer.DacFx" Version="162.1.172" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageReference Include="System.CommandLine.NamingConventionBinder" Version="2.0.0-beta4.22272.1" />
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -12,6 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview1" />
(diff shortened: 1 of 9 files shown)


jmezach (approved):
LGTM. I was just wondering if we need to add a test project for this as well.

ErikEJ:
@jmezach I will add a test project, just need to .Targets in place, maybe you can help?

ErikEJ:
@jmezach I tried and failed, as you can see from a previous commit, maybe you can guide me?

jmezach:
You'll probably want to add a condition on `<CodeAnalysisRulesArgument>` so that it is only set if certain rules are ignored. Right now you're always passing the -ar command line option, but without any arguments.

ErikEJ:
@jmezach > You'll probably want to add a condition on <CodeAnalysisRulesArgument>

😄 - yeah, makes sense, but I have no clue how to do that - sorry!

jmezach:
Something like `<CodeAnalysisRulesArgument Condition="'$(CodeAnalysisRules)'!=''">-ar $(CodeAnalysisRules)</CodeAnalysisRulesArgument>`

jmezach (approved):
LGTM :)

src/DacpacTool/DacpacTool.csproj
```diff
@@ -12,6 +12,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview3" />
```
jeffrosenberg, line 15: Are these actually needed in the csproj file? I don't think I saw any references to them, but let me know if I'm mistaken

jeffrosenberg (approved):
LGTM, with one small comment

src/DacpacTool/DacpacTool.csproj
ErikEJ: They are needed in order to add the additional analyzers .dlls as mentioned in the PR description - I guess that is "runtime" only?

ErikEJ: @jmezach @jeffrosenberg Should I change anything?

jeffrosenberg: If we did an explicit copy, then we probably wouldn't need the `<PackageReference>` at all. I don't have a big problem with using the `<PackageReference>` to copy the files, I just didn't make the connection of what it was doing. At the minimum, I'd probably add a comment clarifying why these are in the project file.

ErikEJ: How about Includeassets="runtime" ?

PR #520, opened 2024-02-27
Add docs of code analysis feature

ErikEJ:
relates to #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,7 +527,28 @@ The database name for the create script gets resolved in the following manner:
 1. Package name.
 > Note: 
 >- the generated script also uses the resolved database name via a setvar command.
->- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`   
+>- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`
+
+## Static code analysis
+Starting with version 2.6.2 of the SDK, there is support for running static code analysis during build. The SDK includes the following sets of rules:
+
+- Microsoft.Rules ([1](https://learn.microsoft.com/previous-versions/visualstudio/visual-studio-2010/dd193411), [2](https://learn.microsoft.com/previous-versions/visualstudio/visual-studio-2010/dd193246) and [3](https://learn.microsoft.com/previous-versions/visualstudio/visual-studio-2010/dd172117))
+- [SqlServer.Rules](https://github.com/tcartwright/SqlServer.Rules/blob/master/docs/table_of_contents.md)
+- [Smells](https://github.com/davebally/TSQL-Smells)
+
+Static code analysis can be enabled by adding the following to the project file:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/2.6.2">
+  <PropertyGroup>
+    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
+    <CodeAnalysisRules>-SqlServer.Rules.SRD0006;-Smells.*</CodeAnalysisRules>
+  </PropertyGroup>
+</Project>
+```
+The optional `CodeAnalysisRules` property allows you to disable individual rules or groups of rules.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -527,7 +527,28 @@ The database name for the create script gets resolved in the following manner:
 1. Package name.
 > Note: 
 >- the generated script also uses the resolved database name via a setvar command.
->- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`   
+>- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`
+
+## Static code analysis
+Starting with version 2.6.2 of the SDK, there is support for running static code analysis during build. The SDK includes the following sets of rules:
```
jmezach, line 533: Maybe we'll want to ship this as 2.7.0 since this is a new feature?

README.md
```diff
@@ -527,7 +527,28 @@ The database name for the create script gets resolved in the following manner:
...
+<Project Sdk="MSBuild.Sdk.SqlProj/2.6.2">
+  <PropertyGroup>
+    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
+    <CodeAnalysisRules>-SqlServer.Rules.SRD0006;-Smells.*</CodeAnalysisRules>
+  </PropertyGroup>
+</Project>
+```
+The optional `CodeAnalysisRules` property allows you to disable individual rules or groups of rules.
+
+A xml file with the analysis results is created in the output folder, and any problems found during analysis are reported as build warnings.
```
jmezach, line 551: Does this work with `TreatWarningsAsErrors`? I think we might want to write something about that here?

jmezach (approved):
Just a couple of comments. Thanks for the docs :).

README.md
ErikEJ: Fine by me - but is it breaking?

README.md
ErikEJ: Let me test

README.md
jmezach: I wouldn't go as far as to make this a 3.0 release no since it is not breaking as you have to opt-in to this feature anyway. But it is more than a simple patch to existing functionality, hence why I think 2.7.0 makes sense in this case.

README.md
jmezach: Perhaps we should call that out in the doc and then add that later if there's a demand for it?

ErikEJ: Added a note

ErikEJ: @jmezach Do you think I should add more?

jmezach: I think this is fine for now.

README.md
```diff
@@ -527,7 +527,30 @@ The database name for the create script gets resolved in the following manner:
 1. Package name.
 > Note: 
 >- the generated script also uses the resolved database name via a setvar command.
->- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`   
+>- if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`
+
+## Static code analysis
+Starting with version 2.7.0 of the SDK, there is support for running static code analysis during build. The SDK includes the following sets of rules:
```
jeffrosenberg, line 533: Can users can also add their own rules by placing them in the correct folder? If so, we should mention that somewhere. If not, you can ignore this! :slightly_smiling_face:

jeffrosenberg (approved):

README.md
ErikEJ: Yeah, it is therotically possible:
- build a strong named rules .dll that targets net6.0 (normally it would target net462)
- copy to the tools folder in the NuGet cache (where Dacpactool.dll is located) - might be complex/not possible on a build server?

But it is not really something I want to support.

I described an alternative approach in the [issue comments](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/170#issuecomment-1962882435), but I want your feedback and user demand first

PR #533, opened 2024-03-07
Add support for "bring your own rules" (or update existing)

### LINKED ISSUE #530: Add ability to update current rules or bring your own
I now find myself fixing bugs to the SqlServer.Rules rule set and having a mechanism to update the ruleset out of band with a SDK update would be handy.

I am proposing the following simple approach, but let me hear what anone else thinks.

- Allow .dll content files to be added to the output folder **possible today**
- Update DacpacTool: If CodeAnalysis is enabled, copy any .dll files from the output folder to the folder of executing assembly
- DacFX Code Analysis will automagically pick up the added / updated rules

Sample .csproj addition to copy .dll files the output folder
```
	<ItemGroup>
	  <None Update="SqlServer.Dac.dll">
	    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
	  </None>
	  <None Update="SqlServer.Rules.dll">
	    <CopyToOutputDirectory>Always</CopyToOutputDirectory>
	  </None>
	</ItemGroup>
```

ErikEJ:
fixes #530


src/DacpacTool/DacpacTool.csproj
```diff
@@ -13,8 +13,8 @@
 
   <ItemGroup>
     <!-- These packages contain DacFX analysis rules, which must be located with the executable in order to be discovered-->
-    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview4" IncludeAssets="runtime" />
-    <PackageReference Include="ErikEJ.DacFX.TSQLSmellSCA" Version="1.0.0-preview1" IncludeAssets="runtime" />
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview7" IncludeAssets="runtime" />
```
jmezach, line 16: Do we now still want to ship these in the box?

jmezach (approved):
LGTM

src/DacpacTool/DacpacTool.csproj
ErikEJ: I think so, they (in particular the SqlServer.Rules) add quite some value - based on experience from production project

src/DacpacTool/PackageAnalyzer.cs
```diff
@@ -21,6 +21,21 @@ public PackageAnalyzer(IConsole console, string rulesExpression)
...
 
+        public void AddRulesFile(FileInfo inputFile)
+        {
+            // Make sure the file exists
+            if (!inputFile.Exists)
+            {
+                throw new ArgumentException($"Unable to find rules file {inputFile}", nameof(inputFile));
+            }
+
+            if (inputFile.Directory.Name.Equals("rules", StringComparison.OrdinalIgnoreCase)
```
jeffrosenberg, line 32: I don't generally like requiring files to be under a certain path, but I'm okay with it in this case, since we don't really have another mechanism. The `Include` elements in the `.csproj` file don't include anything like a `Type` attribute and the `csproj` format isn't within the scope of our project.

src/DacpacTool/PackageBuilder.cs
```diff
@@ -69,6 +69,13 @@ public void AddInputFile(FileInfo inputFile)
                 throw new ArgumentException($"Unable to find input file {inputFile}", nameof(inputFile));
             }
 
+            // Skip custom rules files
```
jeffrosenberg, line 72: It would be helpful to clarify _why_ we're doing this -- i.e. we're copying them to the output folder, and they'll be consumed from there. It's clear in the context of this PR, but in a couple of years we may look back on this and wonder why we'd want to skip certain files.

test/TestProjectWithAnalyzers/TestProjectWithAnalyzers.csproj
```diff
@@ -1,12 +1,17 @@
-<Project>
+﻿<Project>
   <Import Project="$(MSBuildThisFileDirectory)..\..\src\MSBuild.Sdk.SqlProj\Sdk\Sdk.props" />
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
-    <SqlServerVersion>Sql150</SqlServerVersion>
-    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
-    <CodeAnalysisRules>-SqlServer.Rules.SRD0006;-Smells.*;+!SqlServer.Rules.SRN0002</CodeAnalysisRules>      
+      <TargetFramework>netstandard2.0</TargetFramework>
```
jeffrosenberg, line 5: nit: whitespace changes don't seem necessary here.

jeffrosenberg (approved):
Approved, with a couple of small comments

src/DacpacTool/PackageAnalyzer.cs
ErikEJ: Agree, but it was the best solution I could brew up

src/DacpacTool/PackageBuilder.cs
ErikEJ: I added more context to the comment

ErikEJ:
@jeffrosenberg Thanks for your review - I believe I have addressed the comments now

jeffrosenberg (approved):

PR #542, opened 2024-03-22
Fix false positive

### LINKED ISSUE #541: SqlServer.Rules has false positive: SRP0020
https://github.com/tcartwright/SqlServer.Rules/blob/master/docs/Performance/SRP0020.md

```sql
CREATE TABLE [dbo].[OI] (
    [Id]         INT                                         IDENTITY (1, 1) NOT NULL,
    [Version]                   INT                                         NOT NULL,
    CONSTRAINT [PK_OI] PRIMARY KEY NONCLUSTERED ([Id]),
    CONSTRAINT [CI_OI] UNIQUE CLUSTERED ([Version] ASC)
)
```

ErikEJ:
fixes #541

### DIFF
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -13,7 +13,7 @@
 
   <ItemGroup>
     <!-- These packages contain DacFX analysis rules, which must be located with the executable in order to be discovered-->
-    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview8" IncludeAssets="runtime" />
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview9" IncludeAssets="runtime" />
     <PackageReference Include="ErikEJ.DacFX.TSQLSmellSCA" Version="1.0.0-preview2" IncludeAssets="runtime" />
       
     <PackageReference Include="Microsoft.SqlServer.DacFx" Version="162.2.111" />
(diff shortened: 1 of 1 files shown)


jmezach (approved):

src/DacpacTool/DacpacTool.csproj
```diff
@@ -13,7 +13,7 @@
 
   <ItemGroup>
     <!-- These packages contain DacFX analysis rules, which must be located with the executable in order to be discovered-->
-    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0-preview8" IncludeAssets="runtime" />
```
jeffrosenberg, line 16: I feel like this approach of continually incrementing the preview version doesn't scale, especially since we're just rubber-stamping each increment anyway. Could we move to using a wildcard, or possibly a maximum version constraint instead?

jeffrosenberg (approved):

src/DacpacTool/DacpacTool.csproj
ErikEJ: @jeffrosenberg Sure, what would that look like?

ErikEJ: 1.0.* ?

ErikEJ: Or I could release 1.0.0 and we could add the packages to dependabot?

ErikEJ:
PR updated

src/DacpacTool/DacpacTool.csproj
jeffrosenberg: Sorry I didn't respond to this, you had already chosen a path and merged a solution by the time I got to it :slightly_smiling_face: 

In case you haven't seen it, here's the [NuGet Package Version Reference](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort). My recommendation would be to use the "maximum version, exclusive" format (i.e. `(,1.0)`) to set an upper bound at either `1.1.0` or `2.0.0`, and then you're free to release fixes to your package within that range.

PR #543, opened 2024-03-25
pin rules versions

ErikEJ:
fixes #524 PR comments

### DIFF
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -13,8 +13,8 @@
 
   <ItemGroup>
     <!-- These packages contain DacFX analysis rules, which must be located with the executable in order to be discovered-->
-    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0" IncludeAssets="runtime" />
-    <PackageReference Include="ErikEJ.DacFX.TSQLSmellSCA" Version="1.0.0" IncludeAssets="runtime" />
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="[1.0,2.0)" IncludeAssets="runtime" />
+    <PackageReference Include="ErikEJ.DacFX.TSQLSmellSCA" Version="[1.0,2.0)" IncludeAssets="runtime" />
       
     <PackageReference Include="Microsoft.SqlServer.DacFx" Version="162.2.111" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />


src/DacpacTool/DacpacTool.csproj
```diff
@@ -13,8 +13,8 @@
 
   <ItemGroup>
     <!-- These packages contain DacFX analysis rules, which must be located with the executable in order to be discovered-->
-    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="1.0.0" IncludeAssets="runtime" />
-    <PackageReference Include="ErikEJ.DacFX.TSQLSmellSCA" Version="1.0.0" IncludeAssets="runtime" />
+    <PackageReference Include="ErikEJ.DacFX.SqlServer.Rules" Version="[1.0,2.0)" IncludeAssets="runtime" />
```
jeffrosenberg, line 16: Just annotating this for reference: per [this documentation](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort), this format means `1.0 ≤ x < 2.0`

jeffrosenberg (approved):
Thanks for the change!

src/DacpacTool/DacpacTool.csproj
ErikEJ: Thanks for the thoughtful review!  :-)

jmezach:
This does have the side-effect that doing a build today might produce a different result than doing a build tomorrow. This makes it a bit harder to call out any changes to the includes rules in our release notes. Or do we not care about that?

ErikEJ:
@jmezach How about **not** doing this, and consider the rules a dependency like DacFX? (so keep as is) ?

ErikEJ:
I think it makes sense to care about the rule updates, as the will appear as an integral part of the tool going forward

jeffrosenberg:
> I think it makes sense to care about the rule updates, as the will appear as an integral part of the tool going forward
It does make sense, but the challenge is that frequent changes bumping the version number of the dependency don't actually give us anything to review, so it's just a few PRs a week to rubber stamp.

@ErikEJ do these repos serve any purpose outside of this project? Would it possibly make sense to bring them into this repo so that we could review the actual changes themselves?

If you don't want to that -- and there are definitely valid reasons not to -- is there a way to bring these into dependabot or otherwise allow merging to trigger a build without requiring 2 PR reviews?

ErikEJ:
> do these repos serve any purpose outside of this project?

Yes, they do, see my blog post  https://erikej.github.io/dacfx/codeanalysis/sqlserver/2024/04/02/dacfx-codeanalysis.html (and I also plan some VS tools based on the packages)

>  is there a way to bring these into dependabot or otherwise allow merging to trigger a build without requiring 2 PR reviews?

Yeah, it is my understanding that this will happen today, it is just me being pushy 😄

jmezach:
I think these packages should show up in dependabot, at least when there is a stable version, without us having to do anything. I usually merge those PR's without a second reviewer anyway. Might have to look into seeing if that could be automated for dependabot PR's somehow, but that would be a nice to have imho.

Issue #584, opened 2024-07-17
[Idea] .NET Aspire Integration


jmezach:
I just had a completely random thought that perhaps could be interesting. Would it be useful to have some kind of integration with .NET Aspire where one could spin up a SQL Server container and deploy a MSBuild.Sdk.SqlProj project to it.

Definitely interested in what @jeffrosenberg and @ErikEJ think, but if you're a user of this project please let us know if this would be interesting 🤔

ErikEJ:
Yeah, I have been thinking the same, some kind if synergy here seems useful.

jmezach:
Perhaps something like:

```csharp
var databaseProject = builder.AddProject<Projects.MyDatabaseProject>();
var db = builder.AddSqlServer("sql")
                           .AddDatabase("db")
                           .WithDatabaseProject(dacpac)
```

Which would then result in a running SQL Server container named `sql` with a database named `db` and the referenced database project deployed to it.

ErikEJ:
Do you mean WithDatabaseProject(databaseproject) ?

jmezach:
That would be the main API yes. Of course the hard part is figuring out what that is going to do. It would probably involve running sqlpackage somehow. Not sure if that is included in the SQL containers though

ErikEJ:
Sqlpackage is easy to install:

dotnet tool install Microsoft.SqlPackage

With a connection string, would it be needed inside the container? (It only runs on certain Linux distributions)

jmezach:
I guess we could use a locally installed sqlpackage. We can get the connection string from the SQL resource and I'm guessing we can get the path to the dacpac from the project resource. If we can put those together we can make it work.

Might be able to throw some early prototype together on Friday.

ErikEJ:
I have been thinking about ASPIRE0004 and reading about resource references in the apphost.

Maybe what we really want to do is some thing like this?

```csharp
var builder = DistributedApplication.CreateBuilder(args);

builder.AddSqlServer("sql")
            .AddDatabase("test")
            .WithDaccpac("MyDacpac");

builder.Build().Run();
```

The WithDacpac extension method then has some clever ways of finding the location the relevant dacpac and publishing it.

Could be a .dacpac included as content, even pointing to another folder in the solution?

Or a .nupkg with a .dacpac created by MsBuild.SDK .SqlProj

```xml
  <ItemGroup>
    <Content Include="MyDacpac.dacpac">
[...]

PR #585, opened 2024-07-18
Initial .NET Aspire integration prototype

jmezach:
As described in #584 we think there is value in providing some kind of integration between MSBuild.Sdk.SqlProj and .NET Aspire which would essentially allow a developer to spin up a SQL Server container and deploy a database project it. This could significantly enhance the inner loop.

This is a very early prototype of what such an integration could look like. A couple of observations:

- For now I've just build everything into a test .NET Aspire project, but the code in `Extensions.cs` should probably move into a separate project (something like `MSBuild.Sdk.SqlProj.Aspire` maybe?)
- Right now I'm getting a warning ` warning ASPIRE004: '../TestProject/TestProject.csproj' is referenced by an Aspire Host project, but it is not an executable. Did you mean to set IsAspireProjectResource="false"?`. Setting that flag obviously resolves the warning, but then we can't get a hold of the project in `Program.cs`. Perhaps this should be a question to the .NET Aspire team.
- I'm making a lot of assumptions about file paths to figure out the location of the `.dacpac` that needs to be deployed. It looks like .NET Aspire only provides us with the path to the `.csproj` but no other metadata from MSBuild that we can use (at least not that I've seen)
- To deploy the `.dacpac` I'm using the DacFx package directly, so it is essentially part of the AppHost project. Not sure if that's the right way to go, but at least it works.

@ErikEJ @jeffrosenberg Thoughts?


ErikEJ:
@davidfowl This is a really awesome inner loop experience for folks not using EF Core Migrations, but "databasemodel first" or even just Dapper / raw ADO.NET. 

Who can help with the warning mentioned above?

ErikEJ:
@pksorensen FYI! (and any comments / advice you may have)

ErikEJ:
I was able to run and publish the .dacpac!

In order to run this locally in VS, I installed Docker Desktop and updated the .sln file to reference the TestProject. 

I also added some robustness to Extension.cs to avoid some build warnings.

https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/compare/feature/aspire-integration...ErikEJ:aspire-poc?expand=1

jmezach:
I was thinking that maybe the database project should be added as a custom resource, not as the built-in project resource, so you would get something like:

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var databaseProject = builder.AddDatabaseProject<Projects.TestProject>("db");

var sql = builder.AddSqlServer("sql")
                 .AddDatabase("test")
                 .WithDatabaseProject(databaseProject);

builder.Build().Run();
```

Then we can further customize the database project to include additional properties relevant for the deployment as well in the future and scope the `WithDatabaseProject()` method to only accept an `IResourceBuilder<DatabaseProjectResource>` and not just any database project. And we might be able to do some reading from the project file to figure out the right path to the `.dacpac` without hardcoding them.

ErikEJ:
@jmezach Yes, that seems the way to go, and would probably also remove the ASPIRE004 warning

jmezach:
I'm afraid it won't fix that, since that warning is coming from the fact that I've added a reference to the project. If I remove the `AddProject()` call I still get the warning.

It might fix another issue I noticed when I tried to debug my AppHost project. It tried to run and attach a debugger to the database project, which obviously doesn't work.

pksorensen:
I did similar for my database deploy projekt - the problem i ran into was that Aspire do not "create" databases, so the deploy part is responsible of creating the actually database if its missing. This was a gotcha for me that i had to work that out myself. Inspecting the code i dont see anything related to this but maybe bacpac takes care of it

ErikEJ:
@pksorensen Thanks for chiming in! The dacpac magic takes care of creating the database and the objects inside it.

pksorensen:
The goal for me was to make it super easy for a new dev on the project to start the project. So stuff i was looking into was also to be able to run the aspire project (F5 experience) where it restores a database from a pak file to be able to provide initial source data from some "restore point" to make it easier on dev to run/debug issues locally after a project has gone live. (a bit more advanced seed data that is able to fullfill usecases of the project for easier testing)

ErikEJ:
@pksorensen You can also add a SQL MERGE script as a post deploy step to a .dacpac

jmezach:
I've done some more updates on this. First of all I've introduced the `DatabaseProjectResource` type as a first class citizen. This makes it show up on the dashboard as well and I've taken the liberty of pushing logs and updating the deployment status to it as well, see these screenshots:

Screenshot 2024-07-19 at 14 57 29
Screenshot 2024-07-19 at 14 57 34
Screenshot 2024-07-19 at 14 57 49

Now that I'm seeing this I'm wondering if we should perhaps inverse the relation ship between the SqlServer and the database project. So instead of this:

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var databaseProject = builder.AddDatabaseProject<Projects.TestProject>("db");

var sql = builder.AddSqlServer("sql")
                 .AddDatabase("test")
                 .WithDatabaseProject(databaseProject);

builder.Build().Run();
```

If we instead should do this:

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var sql = builder.AddSqlServer("sql")
                 .AddDatabase("test");

var databaseProject = builder.AddDatabaseProject<Projects.TestProject>("db");
                             .DeployTo(sql);

builder.Build().Run();
```

ErikEJ:
Awesome. Would inverting the relationship change the UX?

jmezach:
No, I don't think so :).

ErikEJ:
Agree on the updated fluent syntax!

jmezach:
I made that change now. It also makes the lifecycle hook a bit cleaner as now we only have to loop through the database projects, instead of going through all the SqlServerDatabaseResources and then all the annotation on that.

Have to say I'm pretty happy with how this has turned out so far and I think it can be a real game changer.

ErikEJ:
There are some challenges with finding the .dacpac and ensure the project has been built... Is it possible to use the MsBuild SDK to load the project, or invoke a dotnet build and get some structured output from it?

pksorensen:
I think the DeployTo is interesting. I agree that its easier when reading the code to understand what is going on, but i dont think i seen that in other aspire stuff. Feels like it been mosly "AddReference" and such.

I personally however prefer the `.AddDatabase().WithDatabaseProject()` (to me thats what i would expect from what i have seen on other aspire things).

I am trying to figure out why that "deployTo" does not feel right with me (read, that i dont feel that strongly about it) but i think its something related to that i feel that aspire describes the relationship between resources and not that we are deploying a project onto the database. Deploying is something that is generated from metatada from the relationship metadata in aspire.

Senarie could be that i want to use a database project in the database, but i dont want it to "deploy" it. 

Just thinking out loud, i think this stuff is nice and i can use this as inpiration to my stuff to, thanks @ErikEJ for looping me in.

jmezach:
I tried using `Microsoft.Build` package to fetch the `OutputPath` from the project file by loading it, but then I got errors that it wasn't able to locate the `Microsoft.NET.Sdk`. Perhaps we could resolve that with https://github.com/microsoft/MSBuildLocator.

But I'm not sure whether it is worth it. I think we can safely assume that the project has already been build, since you need to add it as a reference to your AppHost project anyway. As for the "magic" strings, I think those can be assumed as well as we're setting these in the Sdk.props anyway. The only thing I can think of that might provide additional benefit is if we could read the deployment related properties from the project file and use them while deploying the `.dacpac`.

ErikEJ:
"Senarie could be that i want to use a database project in the database, but i dont want it to "deploy" it"

You can not "use" a database project without publishing (deploying) it to a database.

Would PublishTo be easier to understand? (Yeah, naming is hard)

jmezach:
I guess `PublishTo()` would fit well with the current inner loop experience of running `dotnet publish` as well.

ErikEJ:
@jmezach So we could read the properties from just .xml parsing?

ErikEJ:
@jmezach "publish" is also what sqlpackage does

jmezach:
I think we might be able to get to the deployment properties by just using `ProjectRootElement.Open(<path-to-project-file>)` from `Microsoft.Build` package. This doesn't have to evaluate the entire project including all kinds of imports, it just reads the project file as is.

jmezach:
Okay, I gave up and just used MSBuild after all. Now I'm just evaluating the project and getting the `TargetPath` to figure out the path to the `.dacpac` file that needs to be deployed. Works like a charm on my machine at least.

Also renamed the `DeployTo` method to `PublishTo` which I think makes more sense.

ErikEJ:
What if you just included a .dacpac file as content with your app, and would want that deployed?

ErikEJ:
I will have another play with this this weekend. 

Looks like the package name should be something that ends with .Aspire.Hosting

ErikEJ:
Having looked more into Aspire and the configuration patterns, I think this is more natural (but an opition from an actual Aspire expert would be helpful)

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var databaseProject = builder.AddDatabaseProject<Projects.TestProject>("db");

builder.AddSqlServer("sql")
        .AddDatabase("test")
        .WithDatabaseProject(databaseProject);

builder.Build().Run();
```

```csharp
var builder = DistributedApplication.CreateBuilder(args);

builder.AddSqlServer("sql")
        .AddDatabase("test")
        .WithDacpac("myTestDacpac");

builder.Build().Run();
```

jmezach:
One problem I see with that proposal is that in the first case we add a resource to the `DistributedApplicationBuilder` while we don't do that in the latter. Also we're now back to adding things on top of the `SqlServerDatabaseResource` instead of the other way around.

I think having the "dacpac" as a resource in the model kind of makes sense, as we can also hook up logs and status to that resource. Perhaps we should use the "Data-Tier Application" moniker for this resource, so something like this?

Use case - Referenced project:

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var database = builder.AddSqlServer("sql")
        .AddDatabase("test");

var databaseProject = builder.AddDataTierApplication<Projects.TestProject>("db");
        .PublishTo(database);

builder.Build().Run();
```

Use case - Local `.dacpac` file:

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var database = builder.AddSqlServer("sql")
        .AddDatabase("test");

var databaseProject = builder.AddDataTierApplication("db", "/path/to/MyLocal.dacpac");
        .PublishTo(database);

builder.Build().Run();
```

ErikEJ:
> Perhaps we should use the "Data-Tier Application" moniker for this resource, so something like this?

Looks great!

ErikEJ:
Did a quick spike on the other overload:

```csharp
using Aspire.Hosting.Lifecycle;
using Microsoft.Build.Evaluation;
using Microsoft.Build.Locator;
using Microsoft.Extensions.Logging;
using Microsoft.SqlServer.Dac;

namespace Aspire.Hosting;

public static class Extensions
{
    public static IResourceBuilder<DatabaseProjectResource> AddDataTierApplication<TProject>(this IDistributedApplicationBuilder builder, string name)
        where TProject : IProjectMetadata, new()
    {
        MSBuildLocator.RegisterInstance(MSBuildLocator.QueryVisualStudioInstances().OrderByDescending(
            instance => instance.Version).First());

        var resource = new DatabaseProjectResource(name);

        return builder.AddResource(resource)
                      .WithAnnotation(new TProject());
    }

    public static IResourceBuilder<DatabaseProjectResource> AddDataTierApplication(this IDistributedApplicationBuilder builder, string name, string path)
    {
        var resource = new DatabaseProjectResource(name);

        return builder.AddResource(resource)
            .WithAnnotation(new TargetDacpacResourceAnnotation(path));
    }

    public static IResourceBuilder<DatabaseProjectResource> PublishTo(
    this IResourceBuilder<DatabaseProjectResource> builder, IResourceBuilder<SqlServerDatabaseResource> project)
    {
        builder.ApplicationBuilder.Services.TryAddLifecycleHook<DeployDatabaseProjectLifecycleHook>();
        builder.WithAnnotation(new TargetDatabaseResourceAnnotation(project.Resource.Name), ResourceAnnotationMutationBehavior.Replace);
        return builder;
    }

    public static IProjectMetadata? GetProjectMetadata(this DatabaseProjectResource resource)
    {
        return resource.Annotations.OfType<IProjectMetadata>().SingleOrDefault();
    }

    public static TargetDacpacResourceAnnotation? GetDacpacMetadata(this DatabaseProjectResource resource)
    {
        return resource.Annotations.OfType<TargetDacpacResourceAnnotation>().SingleOrDefault();
    }
}

public sealed class DatabaseProjectResource(string name) : Resource(name)
{
    public string GetDacpacPath()
    {
        if (this.GetProjectMetadata() == null)
        {
            if (this.GetDacpacMetadata() is null)
            {
                throw new InvalidOperationException("Path to .dacpac file not specified.");
            }

            return this.GetDacpacMetadata()!.TargetDacpacPath;
        }
        else
        {
            var projectPath = this.GetProjectMetadata()!.ProjectPath;
            var project = new Project(projectPath);
            return project.GetPropertyValue("TargetPath");
        }
    }
}

public record TargetDacpacResourceAnnotation(string TargetDacpacPath) : IResourceAnnotation
{
}
```

ErikEJ:
Adding this to Sdk.Props gets rid of the ASPIRE0004 error, but unsure if it has any other consequences (I cannot see any obvious):

```xml
  <PropertyGroup>
   ....
    <OutputType>Exe</OutputType>
  </PropertyGroup>

```

jmezach:
I guess that would also mean you'd be able to set a MSBuild.Sdk.SqlProj project as the startup project. Not sure what would happen if you'd run it then though.

I can understand the reason for the warning in Aspire if all you can ever do with referenced projects is to run them. But what we want to do is to extract information from the project and do something with that information. Perhaps this warrants a separate discussion over in the Aspire repo?

ErikEJ:
Agree, discussing in the Aspire repo is better than making the project an .exe

ErikEJ:
@jerrynixon Have a look at [this](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/585#issuecomment-2242492669) !

jmezach:
> Agree, discussing in the Aspire repo is better than making the project an .exe

It looks like there's already a similar ask, see https://github.com/dotnet/aspire/issues/3968.

jmezach:
I guess there is one more scenario that would be useful to have here, and that is referencing an MSBuild.Sdk.SqlProj build NuGet package from your AppHost project and deploying that. Main problem with that would be figuring out the path to the `.dacpac` from code. That might require some source generation to extract MSBuild stuff into constants at build time, similar to how the `IProjectMetadata` stuff works for projects. If we are going that route though, we might be able to circumvent the ASPIRE004 warning as well.

ErikEJ:
"one more scenario" - you mean like discussed [here](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/584#issuecomment-2241505103) - agree it would still be useful, although a lot less "inner loopy"

jmezach:
I guess we could leave it for a future version. As a workaround one could add a blank `MSBuild.Sdk.SqlProj` project, add the NuGet package to that and then deploy said project using this feature and I think it would work.

ErikEJ:
NuGet publishes a C# SDK for resolving settings and project references...

jmezach:
@ErikEJ Not entirely sure what you are talking about? ;)

ErikEJ:
@jmezach Sorry, I am referring to: "Main problem with that would be figuring out the path to the .dacpac from code." (from a NuGet PackageReference )

jmezach:
Do you have a link for that?

ErikEJ:
Sure: https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk - I use it here, for example: https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/GUI/Shared/Extensions/ProjectExtensions.cs#L422

jmezach:
Hmm, that would require access to the project file of the host project if I'm not mistaking?

ErikEJ:
@jmezach Yes, to the assets.json file

ErikEJ:
@jmezach I would say that NuGet package support is not required for v1...

jmezach:
Okay, let's see if we can formalize things here a bit.

jmezach:
Right, first question, how are we going to name the package. Suggestions:

1. MSBuild.Sdk.SqlProj.Aspire
1. MSBuild.Sdk.SqlProj.Aspire.Hosting
1. Aspire.Hosting.MSBuildSdkSqlProj

I was initially leaning towards the first. The second one feels a bit weird to me. Third one gels well with the rest of Aspire, but maybe a little too well?

ErikEJ:
@jmezach I vote for number 1, number 3 is not possible, as Aspire is a reserved NuGet name

ErikEJ:
This is what happens in VS when you try to F5 the project as an Exe:

image

And today:
image

jmezach:
I've moved things into a MSBuild.Sdk.SqlProj.Aspire project and cleaned things up a little bit. I've also added support for the local `.dacpac`.

This did reveal another thing we might want to think about. If you add multiple data tier applications to the builder they will be deployed one after the other. Should we maybe do this in parallel?

ErikEJ:
Looks great - getting a lot closer to prime time.

> Should we maybe do this in parallel?

I think it is not a good idea: you will either get blocking or chaos if you do that against the same database, and most likely most scenarios will only inlcude a single .dacpac (adn we support multiple in series)

The seperate project could have some Nuget magic sprinkled over it, (like source link etc) - but we can postpone that to a seperate PR.

The advise from Aspire is to version components the same as the Aspire version they refer - something like 8.0.2-preview.1

ErikEJ:
So Aspire 8.1 is out now...

src/MSBuild.Sdk.SqlProj.Aspire/DataTierApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+using Aspire.Hosting.Lifecycle;
+using Microsoft.Build.Locator;
+using MSBuild.Sdk.SqlProj.Aspire;
+
+namespace Aspire.Hosting;
+
+public static class DataTierApplicationBuilderExtensions
+{
+    /// <summary>
+    /// Adds a data-tier application resource to the application based on a referenced project.
```
ErikEJ, line 11: Maybe: "Adds a data-tier application resource to the application based on a referenced MSBuild.Sdk.SqlProj project." ?

src/MSBuild.Sdk.SqlProj.Aspire/DataTierApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </summary>
+    /// <typeparam name="TProject">Type that represents the project that produces the data-tier application package.</typeparam>
```
ErikEJ, line 13: Maybe: Type that represents the project that produces the data-tier application package (.dacpac file).

src/MSBuild.Sdk.SqlProj.Aspire/DataTierApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        var resource = new DataTierApplicationResource(name);
+        
+        return builder.AddResource(resource)
+                      .WithAnnotation(new TProject());
+    }
+
+    /// <summary>
+    /// Adds a data-tier application resource to the application based on a data-tier application package file.
```
ErikEJ, line 33: Maybe: Adds a data-tier application resource to the application based on a data-tier application package file (.dacpac).

src/MSBuild.Sdk.SqlProj.Aspire/DataTierApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// </summary>
+    /// <param name="builder">An <see cref="IDistributedApplicationBuilder"/> instance to add the data-tier application to.</param>
+    /// <param name="name">Name of the resource.</param>
+    /// <param name="dacpacPath">Path to a data-tier application package file.</param>
```
ErikEJ, line 37: Maybe: Path to a data-tier application package file (.dacpac).

jmezach:
According to this Sourcelink should be enabled by default: https://github.com/dotnet/sourcelink/blob/main/README.md#using-source-link-in-net-projects

ErikEJ:
Re Sourcelink: missed this doc, I guess I was an early adopter.  Thanks

src/MSBuild.Sdk.SqlProj.Aspire/MSBuild.Sdk.SqlProj.Aspire.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsPackable>true</IsPackable>
+    <PackageTags>aspire hosting sqlserver sql dacfx sqlproj</PackageTags>
+    <Description>.NET Aspire integration for MSBuild.Sdk.SqlProj projects</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting.SqlServer" Version="8.0.2" />
```
ErikEJ, line 13: "8.1.0" is out now

test/TestAspireHost/TestAspireHost.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <IsAspireHost>true</IsAspireHost>
+    <UserSecretsId>e143fa64-20a8-42e1-98a2-c5d209a01e80</UserSecretsId>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Aspire.Hosting.AppHost" Version="8.0.2" />
```
ErikEJ, line 13: "8.1.0" is out now

test/TestAspireHost/TestAspireHost.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageReference Include="Aspire.Hosting.SqlServer" Version="8.0.2" />
```
ErikEJ, line 14: "8.1.0"

ErikEJ:
LGTM, only thing missing before a preview is a readme for the NuGet package.

```xml
<PropertyGroup>
   ...
   <PackageReadmeFile>readme.md</PackageReadmeFile>
  ...
</PropertyGroup>

<ItemGroup>
       <None Include="readme.md" Link="readme.md" Pack="true" PackagePath="\" />
</ItemGroup>
```

jmezach:
I have a preview package now. I can manually upload that to NuGet.org so people can try it out?

ErikEJ:
Looking forward to getting feedback on the preview - I will make some noise about it!

ErikEJ (approved):

jmezach:
Package is up: https://www.nuget.org/packages/MSBuild.Sdk.SqlProj.Aspire/8.1.0-preview-0003-g5786f49f16

I'm thinking we might want to add some tests as well?

ErikEJ:
Yeah, some kind of testing would be good... (Other than smoke testing)

ErikEJ:
Could update the smoke test to use the published package?

`    <PackageReference Include="MSBuild.Sdk.SqlProj.Aspire" Version="8.1.0-preview-0003-g5786f49f16" />`

nit: This is missing from the package:

```xml
	<PackageProjectUrl>https://github.com/rr-wfm/MSBuild.Sdk.SqlProj</PackageProjectUrl>
	<PublishRepositoryUrl>true</PublishRepositoryUrl>

```

ErikEJ:
I have the 8.400 SDK preview installed, and this causes the MsBuildLocator to pick the wrong SDK, causing runtime errors.

Changing to this fixed it for me (it now picks 8.0.303)

```
            var instances = MSBuildLocator.QueryVisualStudioInstances();
            MSBuildLocator.RegisterInstance(instances.First());

```

ErikEJ:
Or maybe just:

`MSBuildLocator.RegisterDefaults();`

ErikEJ:
@jmezach Sorry I did not smoke test with the recent changes using MsBuildLocator 😢

jmezach:
Just uploaded a new version to NuGet.org which contains these fixes. Can you give that a spin? https://www.nuget.org/packages/MSBuild.Sdk.SqlProj.Aspire/8.1.0-preview-0004-gcf7cb42ce4. It seemed to work fine on my box.

ErikEJ:
"Works on my PC" too 🎉

ErikEJ:
https://x.com/ErikEJ/status/1816125223667917134

ErikEJ:
Proposal from X for naming:

` .AddSqlProject<ProjectType>(projectName)`
and 
`.AddSqlProject(projectName).FromDacPac(file name)`

jmezach:
Yeah, I guess the data-tier application naming didn't go down so well ;). I'd be more than happy to change it to this.

ErikEJ:
Looks like this is still nedded set to make the NuGet package pass all validations in NuGet package explorer:

`<DebugType>embedded</DebugType>`

and 

<PropertyGroup Condition="'$(GITHUB_ACTIONS)' == 'true'">
  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
</PropertyGroup>

https://blog.elmah.io/validate-nuget-packages-before-publishing-from-github-actions/

image

ErikEJ:
>  I'd be more than happy to change it to this.

I think it is a good suggestion, yes

ErikEJ:
Looks great! @robertmclaws FYI

ErikEJ:
And package looks good now!

image

jmezach:
Cool, I'll upload the new package to NuGet.org as well.

.github/workflows/main.yml
```diff
@@ -70,13 +65,24 @@ jobs:
...
+
     # Upload SDK package
     - name: upload SDK
       uses: actions/upload-artifact@v4
       with:
         name: sdk-package
         path: ./src/MSBuild.Sdk.SqlProj/bin/Release/
     
+    # Upload Aspire package
+    - name: upload SDK
```
ErikEJ, line 80: -name: upload Aspire

ErikEJ:
Observation: I still need a global.json like this, to make the MsBuildLocator work when I have a preview SDK installed:

```json
{
  "sdk": {
    "version": "8.0.303",
    "allowPrerelease": false,
    "rollForward": "disable"
  }
}

```

jmezach:
What kind of error do you get when you don't have that global.json?

ErikEJ:
Proposal for a slightly better UX (no "Unknown" status displayed) - in particular with multiple sql projects

```csharp
using Aspire.Hosting.ApplicationModel;
using Aspire.Hosting.Lifecycle;
using Microsoft.Extensions.Logging;
using Microsoft.SqlServer.Dac;

namespace MSBuild.Sdk.SqlProj.Aspire;

public class PublishSqlProjectLifecycleHook : IDistributedApplicationLifecycleHook
{
    private readonly ResourceLoggerService _resourceLoggerService;
    private readonly ResourceNotificationService _resourceNotificationService;

    public PublishSqlProjectLifecycleHook(ResourceLoggerService resourceLoggerService,
        ResourceNotificationService resourceNotificationService)
    {
        _resourceLoggerService = resourceLoggerService ?? throw new ArgumentNullException(nameof(resourceLoggerService));
        _resourceNotificationService = resourceNotificationService ?? throw new ArgumentNullException(nameof(resourceNotificationService));
    }

    public async Task AfterResourcesCreatedAsync(DistributedApplicationModel application, CancellationToken cancellationToken)
    {
        var validItems = new List<(SqlProjectResource SqlProject, string DacpacPath)>();

        foreach (var sqlProject in application.Resources.OfType<SqlProjectResource>())
        {
            var logger = _resourceLoggerService.GetLogger(sqlProject);

            var dacpacPath = sqlProject.GetDacpacPath();
            if (!File.Exists(dacpacPath))
            {
                logger.LogError("SQL Server Database project package not found at path {DacpacPath}.", dacpacPath);
                await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                    state => state with { State = new ResourceStateSnapshot("Failed", KnownResourceStateStyles.Error) });
                continue;
            }

            validItems.Add((sqlProject, dacpacPath));

            await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                state => state with { State = new ResourceStateSnapshot("Loading", KnownResourceStateStyles.Info) });
        }

        foreach (var item in validItems)
        {
            var logger = _resourceLoggerService.GetLogger(item.SqlProject);

            var dacpacPath = item.DacpacPath;

            var targetDatabaseResourceName = item.SqlProject.Annotations.OfType<TargetDatabaseResourceAnnotation>().Single().TargetDatabaseResourceName;
            var targetDatabaseResource = application.Resources.OfType<SqlServerDatabaseResource>().Single(r => r.Name == targetDatabaseResourceName);
            var connectionString = await targetDatabaseResource.ConnectionStringExpression.GetValueAsync(cancellationToken);


            try
            {
                var dacServices = new DacServices(connectionString);
                dacServices.Message += (sender, args) => logger.LogInformation(args.Message.ToString());

                var dacpacPackage = DacPackage.Load(dacpacPath, DacSchemaModelStorageType.Memory);

                await _resourceNotificationService.PublishUpdateAsync(item.SqlProject,
                    state => state with { State = new ResourceStateSnapshot("Publishing", KnownResourceStateStyles.Info) });

                dacServices.Deploy(dacpacPackage, targetDatabaseResource.Name, true, new DacDeployOptions(), cancellationToken);

                await _resourceNotificationService.PublishUpdateAsync(item.SqlProject,
                    state => state with { State = new ResourceStateSnapshot("Published", KnownResourceStateStyles.Success) });
            }
            catch (Exception ex)
            {
                logger.LogError(ex, "Failed to publish database project.");

                await _resourceNotificationService.PublishUpdateAsync(item.SqlProject,
                    state => state with { State = new ResourceStateSnapshot("Failed", KnownResourceStateStyles.Error) });
            }
        }
    }
}

```

ErikEJ:
> What kind of error do you get when you don't have that global.json?


```
System.AggregateException
  HResult=0x80131500
  Message=One or more errors occurred. (Invalid static method invocation syntax: "[MSBuild]::SubstringByAsciiChars($(MSBuildProjectFile), 0, 8)". Method '[MSBuild]::SubstringByAsciiChars' not found. Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)). Check that all parameters are defined, are of the correct type, and are specified in the right order.  C:\Program Files\dotnet\sdk\8.0.400-preview.0.24324.5\Microsoft.Common.CurrentVersion.targets)
  Source=System.Private.CoreLib
  StackTrace:
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at Aspire.Hosting.DistributedApplication.Run() in /_/src/Aspire.Hosting/DistributedApplication.cs:line 338
   at Program.<Main>$(String[] args) in C:\Code\Github\MSBuild.Sdk.SqlProj\test\TestAspireHost\Program.cs:line 13

  This exception was originally thrown at this call stack:
    Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject(string, Microsoft.Build.Shared.IElementLocation, string, object[])
    Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject<T1, T2>(Microsoft.Build.Shared.IElementLocation, string, T1, T2)
    Microsoft.Build.Evaluation.Expander<P, I>.Function<T>.Execute(object, Microsoft.Build.Evaluation.IPropertyProvider<T>, Microsoft.Build.Evaluation.ExpanderOptions, Microsoft.Build.Shared.IElementLocation)
    Microsoft.Build.Evaluation.Expander<P, I>.PropertyExpander<T>.ExpandPropertyBody(string, object, Microsoft.Build.Evaluation.IPropertyProvider<T>, Microsoft.Build.Evaluation.ExpanderOptions, Microsoft.Build.Shared.IElementLocation, Microsoft.Build.Evaluation.UsedUninitializedProperties, Microsoft.Build.Shared.FileSystem.IFileSystem)
    Microsoft.Build.Evaluation.Expander<P, I>.PropertyExpander<T>.ExpandPropertiesLeaveTypedAndEscaped(string, Microsoft.Build.Evaluation.IPropertyProvider<T>, Microsoft.Build.Evaluation.ExpanderOptions, Microsoft.Build.Shared.IElementLocation, Microsoft.Build.Evaluation.UsedUninitializedProperties, Microsoft.Build.Shared.FileSystem.IFileSystem, Microsoft.Build.BackEnd.Logging.LoggingContext)
    Microsoft.Build.Evaluation.Expander<P, I>.PropertyExpander<T>.ExpandPropertiesLeaveEscaped(string, Microsoft.Build.Evaluation.IPropertyProvider<T>, Microsoft.Build.Evaluation.ExpanderOptions, Microsoft.Build.Shared.IElementLocation, Microsoft.Build.Evaluation.UsedUninitializedProperties, Microsoft.Build.Shared.FileSystem.IFileSystem, Microsoft.Build.BackEnd.Logging.LoggingContext)
    Microsoft.Build.Evaluation.Expander<P, I>.ExpandIntoStringLeaveEscaped(string, Microsoft.Build.Evaluation.ExpanderOptions, Microsoft.Build.Shared.IElementLocation, Microsoft.Build.BackEnd.Logging.LoggingContext)
    Microsoft.Build.Evaluation.Evaluator<P, I, M, D>.EvaluatePropertyElement(Microsoft.Build.Construction.ProjectPropertyElement)
    Microsoft.Build.Evaluation.Evaluator<P, I, M, D>.EvaluatePropertyGroupElement(Microsoft.Build.Construction.ProjectPropertyGroupElement)
    Microsoft.Build.Evaluation.Evaluator<P, I, M, D>.PerformDepthFirstPass(Microsoft.Build.Construction.ProjectRootElement)
    ...
    [Call Stack Truncated]

Inner Exception 1:
InvalidProjectFileException: Invalid static method invocation syntax: "[MSBuild]::SubstringByAsciiChars($(MSBuildProjectFile), 0, 8)". Method '[MSBuild]::SubstringByAsciiChars' not found. Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)). Check that all parameters are defined, are of the correct type, and are specified in the right order.  C:\Program Files\dotnet\sdk\8.0.400-preview.0.24324.5\Microsoft.Common.CurrentVersion.targets

```

jmezach:
That looks like an issue in the preview SDK to me. Can you build the MSBuild.Sdk.SqlProj project using the preview SDK just fine?

ErikEJ:
I cannot! Getting errors about other "Missing Method" exceptions - that preview SDK looks broken.

ErikEJ:
More info here - https://github.com/davidfowl/WaitForDependenciesAspire

ErikEJ:
Given all the hoops even with .dacpac deployment, it could just be an unsupported scenario. But using the healthcheck to trigger deployment seems useful.

jmezach:
Yeah, I think this whole waiting for dependencies thing requires some built-in support in .NET Aspire. There is an open issue for that here: https://github.com/dotnet/aspire/issues/921, although it seems they aren't willing to built that in.

I don't think there is a more appropriate hook that we could use for this though, do you?

ErikEJ:
> I don't think there is a more appropriate hook that we could use for this though, do you?

No, tried various, to no avail. So spinning up a Server side and a client side and have everything ready is a challenge, I think we must wait for the Aspire team to address that better (WaitFor is not even in the product yet)

jmezach:
Does `WaitFor` work with our stuff though? As in, would that solve anything?

ErikEJ:
No, for our purpose it does not make any difference, ideally we probably could have some ResourceReady event based on a Health Check and fire on that (if applicable)

jmezach:
Yeah. I think another issue with this right now is that the health check isn't going to be healthy until the database has actually been created. But it won't be created until the dacpac has been deployed, so it is a bit of a chicken and egg problem there. I guess you could work around that by having the health check at the server level and not the database level, but I'm not sure what that would look like. But maybe we should leave this for now and wait until Aspire has something built-in for this.

I've been adding some unit tests, but ran into a bit of a challenge with the actual deployment stuff as we currently don't have a way to mock out the actual calls to `DacServices`. Might need to put a small abstraction around that to aid with testing.

ErikEJ:
> But maybe we should leave this for now and wait until Aspire has something built-in for this. 

Agree!

jmezach:
I've been looking into these tests and created an abstraction now. However, now I'm running into an issue that in my tests there is no tcp endpoint allocated for the SQL Server resources, which I guess makes sense. But I'm not entirely sure how to fix this. I also couldn't find any existing tests in the Aspire repo that do this.

ErikEJ:
Maybe the testing story is not fully baked yet?

jmezach:
Nevermind, I figured something out by simply adding an explicit endpoint in the test :). Already found another interesting bug where apparently the MSBuild projects were being cached somewhere, causing a subsequent attempt to fetch the dacpac path from it to fail.

jmezach:
I got a basic test for the lifecycle hook working now. I was kinda hoping to also test the interaction with the `ResourceNotificationService` and the `ResourceLoggerService` as well, but since these are classes not interfaces it is a bit hard to mock out right now.

ErikEJ:
My colleague @pksorensen has a prototype of a solution for the waiting issue  - he will share on Github soon

pksorensen:
@ErikEJ 

The code i used for specifying needs/wait from resources to depending resources is here:
https://github.com/EAVFW/EAVFramework/tree/pks/tpc/aspire/EAVFramework.Extensions.Aspire.Hosting

My project is very similar to your project.
```
using Aspire.Hosting;
using EAVFramework.Extensions.Aspire.Hosting;

var builder = DistributedApplication.CreateBuilder(args);

var sqlserver = builder.AddSqlServer("MCAspireSQL");

var db = sqlserver.AddDatabase("MC");

var eavmodel = builder.AddEAVFWModel<Projects.MC_Models>("mc-model")
    .WithNetTopologySuite()
    .PublishTo(db);

var management = builder.AddProject<Projects.MC_Portal>("mc-admin");

management
    .WithEnvironment("EAVFW_JOBSERVER_DISABLED","true")
    .WithEnvironment("SubscriberQueue__Disabled","true")
    .WithReference(db, "ApplicationDB")
    .Needs(eavmodel);


builder.Build().Run();
```

That `.Needs(eavmodel)` tells that the project mc-admin should not spin up before eavmodel is in the "Running" state. I took the code from fowler and modified it abit. I did consider to lock it further down to just my resources but left it like this for now. If his spike becomes a aspire package at somepoint one might remove that code again.

I considered removing the health check part as it dont really provide any value given i can control the state and when putting it to "Running" i know its running so no need to do a extra "healthcheck"
https://github.com/EAVFW/EAVFramework/blob/pks/tpc/aspire/EAVFramework.Extensions.Aspire.Hosting/AspireEAVFWHealthCheck.cs

Significant change from what you have here is that i run the lifecycle at start and then i try with backup waiting for the resources (DB) to come up.
https://github.com/EAVFW/EAVFramework/blob/pks/tpc/aspire/EAVFramework.Extensions.Aspire.Hosting/PublishEAVFWProjectLifecycleHook.cs#L47

This was due to if we do it in the aftercreated it wont be fired due to the needs/wait for logic wont allow app to come to the created state. 

image
image

image

ErikEJ:
Could Fowlers WaitFor be used instead of your Needs? @pksorensen 

Seems the main change is that you intercept the kifecycle early and then wait for the db resource to be ready.. and avoid blocking the wait for ( as we currently do)

pksorensen:
it is fowlers wait stuff i used, i just renamed it to needs. 

Anyway, you can use it directly as is because if you adds wait on your project, it wont ever come to the resource created life cycle hook. So you have to kick it off as i do in the start phase and have a task running until the db is ready and then apply the db.

Fowler is also kicking off a long running task watching state changes until the dependency gets to the running state.

ErikEJ:
@jmezach Looks like the current state of the PR prevents running the apphost, getting timeout errors

ErikEJ:
@jmezach I tried to adapt the apporach from @pksorensen (just a POC):

```csharp
using Aspire.Hosting.ApplicationModel;

namespace MSBuild.Sdk.SqlProj.Aspire
{
    public record ServerReadyAnnotation() : IResourceAnnotation
    {
        public int Attempt { get; set; } = 1;
        public bool Success { get; set; }
        public bool Published { get; set; }
    }
}
```


```csharp
using Aspire.Hosting.ApplicationModel;
using Aspire.Hosting.Lifecycle;
using Microsoft.Data.SqlClient;
using Microsoft.Extensions.Logging;

namespace MSBuild.Sdk.SqlProj.Aspire;

internal sealed class PublishSqlProjectLifecycleHook 
    : IDistributedApplicationLifecycleHook, IAsyncDisposable
{
    private readonly IDacpacDeployer _deployer;
    private readonly ResourceLoggerService _resourceLoggerService;
    private readonly ResourceNotificationService _resourceNotificationService;
    private readonly CancellationTokenSource _tokenSource = new();

    public PublishSqlProjectLifecycleHook(
        IDacpacDeployer deployer, 
        ResourceLoggerService resourceLoggerService,
        ResourceNotificationService resourceNotificationService)
    {
        _deployer = deployer ?? throw new ArgumentNullException(nameof(deployer));
        _resourceLoggerService = resourceLoggerService ?? throw new ArgumentNullException(nameof(resourceLoggerService));
        _resourceNotificationService = resourceNotificationService ?? throw new ArgumentNullException(nameof(resourceNotificationService));
    }

    public Task BeforeStartAsync(DistributedApplicationModel application, CancellationToken cancellationToken = default)
    {
        foreach (var resource in application.Resources.OfType<SqlProjectResource>())
        {
            StartDacpacDeployment(resource, application, _tokenSource.Token);
        }

        return Task.CompletedTask;
    }

    private void StartDacpacDeployment(SqlProjectResource sqlProject, DistributedApplicationModel application, CancellationToken cancellationToken)
    {
        var logger = _resourceLoggerService.GetLogger(sqlProject);

        _ = Task.Run(async () =>
        {
            int retryCount = 10;
            int delayInMillisecodns = 1000;

            for (int i = 0; i < retryCount; i++)
            {
                try
                {
                    var dacpacPath = sqlProject.GetDacpacPath();
                    if (!File.Exists(dacpacPath))
                    {
                        logger.LogError("SQL Server Database project package not found at path {DacpacPath}.", dacpacPath);
                        await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                            state => state with { State = new ResourceStateSnapshot(KnownResourceStates.FailedToStart, KnownResourceStateStyles.Error) });
                        continue;
                    }

                    var targetDatabaseResourceName = sqlProject.Annotations.OfType<TargetDatabaseResourceAnnotation>().Single().TargetDatabaseResourceName;
                    var targetDatabaseResource = application.Resources.OfType<SqlServerDatabaseResource>().Single(r => r.Name == targetDatabaseResourceName);
                    var targetServerResource = targetDatabaseResource.Parent;
                    var connectionString = await targetDatabaseResource.ConnectionStringExpression.GetValueAsync(cancellationToken);
                    if (connectionString == null)
                    {
                        logger.LogError($"Failed to retrieve connection string for target database {targetDatabaseResourceName}.", targetDatabaseResourceName);
                        await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                            state => state with { State = new ResourceStateSnapshot(KnownResourceStates.FailedToStart, KnownResourceStateStyles.Error) });
                        continue;
                    }

                    if (!sqlProject.TryGetLastAnnotation(out ServerReadyAnnotation? serverReadyAnnotation))
                    {
                        sqlProject.Annotations.Add(serverReadyAnnotation = new ServerReadyAnnotation());
                    }

                    if (!serverReadyAnnotation.Success)
                    {
                        await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                            state => state with { State = new ResourceStateSnapshot($"Checking server readiness - Attempt {serverReadyAnnotation.Attempt}", KnownResourceStateStyles.Info) });

                        try
                        {
                            var serverConnectionString = await targetServerResource.ConnectionStringExpression.GetValueAsync(CancellationToken.None);

                            using var conn = new SqlConnection(serverConnectionString);
                            await conn.OpenAsync();

                            using var cmd = conn.CreateCommand();
                            cmd.CommandText = "SELECT 1";
                            await cmd.ExecuteScalarAsync();

                            await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                                state => state with { State = new ResourceStateSnapshot($"Server ready", KnownResourceStateStyles.Info) });

                            serverReadyAnnotation.Success = true;
                        }
                        catch (InvalidOperationException invalid)
                        {
                            serverReadyAnnotation.Attempt++;
                            logger.LogWarning(invalid, "Transient error, probably due to endpoints not up yet. Backing off to retry.");
                            throw;
                        }
                        catch (SqlException sqlexception)
                        {
                            serverReadyAnnotation.Attempt++;
                            logger.LogWarning(sqlexception, "Transient error, probably due to sql server not ready yet. Backing off to retry.");
                            throw;
                        }
                        catch (Exception ex)
                        {
                            logger.LogError(ex, "Failed to contact server");

                            await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                                state => state with { State = new ResourceStateSnapshot(KnownResourceStates.FailedToStart, KnownResourceStateStyles.Error) });

                            throw;
                        }
                    }

                    await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                        state => state with { State = new ResourceStateSnapshot("Publishing", KnownResourceStateStyles.Info) });

                    try
                    {
                        if (!serverReadyAnnotation.Published)
                        {
                            _deployer.Deploy(dacpacPath, connectionString, targetDatabaseResource.DatabaseName, logger, cancellationToken);
                            serverReadyAnnotation.Published = true;
                        }

                        await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                                state => state with { State = new ResourceStateSnapshot(KnownResourceStates.Running, KnownResourceStateStyles.Success) });
                    }
                    catch (Exception ex)
                    {
                        logger.LogError(ex, "Failed to publish database project.");

                        await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                            state => state with { State = new ResourceStateSnapshot(KnownResourceStates.FailedToStart, KnownResourceStateStyles.Error) });
                    }
                }
                catch
                {
                    if (i == retryCount - 1)
                    {
                        await _resourceNotificationService.PublishUpdateAsync(sqlProject,
                            state => state with { State = new ResourceStateSnapshot(KnownResourceStates.FailedToStart, KnownResourceStateStyles.Error) });
                    }
                    else
                    {
                        await Task.Delay(delayInMillisecodns);
                        delayInMillisecodns *= 2; // Exponential backoff
                    }
                }
            }
        }, cancellationToken);
    }

    public ValueTask DisposeAsync()
    {
        _tokenSource.Cancel();
        return default;
    }

    public Task AfterResourcesCreatedAsync(DistributedApplicationModel application, CancellationToken cancellationToken)
    {
        return Task.CompletedTask;
    }
}

```

ErikEJ:
@jmezach I now got this advanced scenario running, where an app that needs the schema to be present during startup (the Data API Builder container app) waits for the .dacpac publish to complete before proceeding. Based on a simplified version of the waiting pattern from Fowler shared by @pksorensen 

Not sure what the best way to share my spike is? One option could be to add what we have now to master, and I could do a couple of PRs based on that? 

Anyway, I have a commit with my changes here: https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/commit/638f98896ca026d6c6af34cd17cdc4e73e5d04f7 - that broke the current tests :-( 

```csharp
var builder = DistributedApplication.CreateBuilder(args);

var sqlPassword = builder.AddParameter("sql-password");
var sqlServer = builder
    .AddSqlServer("sql", sqlPassword, port: 1234)
    .WithDataVolume("MyDataVolume");

var sqlDatabase = sqlServer.AddDatabase("Database");

var sqlProject = builder.AddSqlProject<Projects.TestProject>("testproject")
    .PublishTo(sqlDatabase);

var dabConfig = "./dab-config.json";
var dabServer = builder
    .AddContainer("data-api", "mcr.microsoft.com/azure-databases/data-api-builder")
    .WithBindMount(dabConfig, "/App/dab-config.json")
    .WithHttpEndpoint(port: 5000, targetPort: 5000, name: "http")
    .WithReference(sqlDatabase)
    .WaitForSqlProject(sqlProject);

builder.Build().Run();
```

jmezach:
I guess the question here is whether we think that the current implementation has value on its own. I'd rather nog merge something that doesn't work in real-life scenario's.

ErikEJ:
I think for real world scenarios there will be apps starting and expecting the database objects to be available. I believe the approach proposed in my branch solves that.

pksorensen:
A bit of learning from playing with my stuff was that its properly a good idea also to test/play abit with the aspir8 cli and deployment/running connected to an azure connection. I found out that those tools have some dependency on the manifest that aspire generates and some of my stuff broke that and i had to add some "ignoreFromManifest" on my stuff. 

I dont have a userbase on my toy project so i just release to playaround, but if there is a somewhat userbase that want to play with it, they might run into issues. but thats also away to get the feedback i guess :D 

I see alot of value from what i have done with mine and its very similar to what you guys are doing - that we are able to create nice f5 experience ind dev for our soltuions.  As a new dev on a project where you have to go find out how the architecture is and setting up a database / publish its really awesome if there is a aspire app host and one just run it knowing it sets up everything in dev is a huge win for me.  I am not there yet that i feeld the hole cloud/deployment experience is working (atleast for our enterprise projects)

so if written what the expectations are for users that sees it and want to try it, i do think its good to get things out there.

pksorensen:
Its my understanding that without your changes it will not successfully deploy every time and as a user that would be annoying if you have to run it again to get it working. It gives the best experience that it just work and deploys, even if we feel that its a bit a hack to start task and wait for it to get ready. So ye i think your stuff should be included to give the best experience of people trying it

jmezach:
@ErikEJ Feel free to merge in your changes to this PR. I'm swamped tomorrow, but I can have a look at it on Friday probably

ErikEJ:
@jmezach Could not figure out how to merge, but I have added a PR against your feature branch

ErikEJ:
Fine with merging this as is, will work well in many use cases.

ErikEJ:
This looks promising: 

https://github.com/dotnet/aspire/issues/5234

ErikEJ:
https://github.com/dotnet/aspire/issues/5501

Thompson1985 (approved):

ErikEJ:
@jmezach https://www.youtube.com/live/ujYBzEEuir0?si=968QKmAsvZ-ViNzm&t=9220 😄

Thompson1985 (approved):

jmezach:
I've done some refactoring here to make this compatible with .NET Aspire 9. It now waits for the SqlDatabaseResource to become ready and that seems to work quite well on my machine. @ErikEJ Would you mind taking it for a spin as well?

There's still a bit of work to do here. One is fixing the unit tests. And I think we might want to be publishing events ourselves as well so that folks could potentially create dependencies between `SqlProjectResource`s if they desire.

ErikEJ:
So PublishTo will now do the waiting for the ReadyEvent, and not publish until the database container is ready to accept connections?

Some thoughts on future work after 1.0:

- support projects with references
- support .sqlproj and/or .sqlprojx
- sanity check that the linked project is a valid project?

ErikEJ:
trying classic .sqlproj:

Exception thrown: 'Microsoft.Build.Exceptions.InvalidProjectFileException' in Microsoft.Build.dll

(Maybe we should add some error handling to better report this?)

(I think .sqlproj is not a high priority)

trying .sqlprojx:

TargetPath returns: C:\Code\Github\MSBuild.Sdk.SqlProj\test\Test\bin\Debug\Database2.dll (so close, maybe we can compensate?)

(I think .sqlprojx support would be really nice)

jmezach:
I'm guessing .sqlprojx also produces a `.dll` because it support SQLCLR so the `TargetPath` probably points at the SQLCLR assembly. I guess we could replace `.dll` with `.dacpac` but that is a bit of an assumption.

As for the better error reporting, where are you seeing that error currently?

ErikEJ:
.sqlprojx has a `SqlTargetPath` !! (that points to the .dacpac file)

Error/crash seen in GetDacpacPath() when trying to load a classic .sqlproj

jmezach:
Made some changes so that we try `SqlTargetPath` first. If that's not available we try `TargetPath`.

Also improved the error handling a bit.

ErikEJ:
Publishing ready event would be a great addition, so we can support scenarios like Data Api Builder expecting the schema to be present at startup.

ErikEJ:
@jmezach Can confirm .sqlprojx support works great!

ErikEJ:
Just saw this: https://github.com/CommunityToolkit/Aspire

jmezach:
@ErikEJ Yeah, saw that too. Looks interesting. Not entirely sure though if it makes sense to move this to the toolkit. Or do you feel like it should be part of it?

ErikEJ:
@jmezach There are pros and cons, but it looks like we will gain a lot of visibility in VS https://github.com/CommunityToolkit/Aspire/blob/main/docs/faq.md#finding-community-toolkit-integrations 

Con: More formal constrains, but does not seem too bad.

ErikEJ:
This is what the integration / visibility looks like:

image

jmezach:
> This is what the integration / visibility looks like:
> 
> image

Those are just the "official" packages from the Aspire project I think?

ErikEJ:
I think the view will also include packages from community according to the FAQ linked above..

ErikEJ:
Full error message when trying to load a classic .sqlproj:

```
Microsoft.Build.Exceptions.InvalidProjectFileException
  HResult=0x80131500
  Message=The imported file "$(MSBuildExtensionsPath)\Microsoft\VisualStudio\v$(VisualStudioVersion)\SSDT\Microsoft.Data.Tools.Schema.SqlTasks.targets" does not exist and appears to be part of a Visual Studio component. This file may require MSBuild.exe in order to be imported successfully, and so may fail to build in the dotnet CLI.   C:\Code\Github\MSBuild.Sdk.SqlProj\test\Database\Database.sqlproj
  Source=Microsoft.Build
```

Wonder if we could do some probing given we have the path to the .sqlproj?

ErikEJ:
I managed to add support for classic .sqlproj with the following changes:

```csharp
using Aspire.Hosting;
using Aspire.Hosting.ApplicationModel;
using Microsoft.Build.Evaluation;
using Microsoft.Build.Exceptions;

namespace MSBuild.Sdk.SqlProj.Aspire;

public sealed class SqlProjectResource(string name) : Resource(name)
{
    public string GetDacpacPath()
    {
        var projectMetadata = Annotations.OfType<IProjectMetadata>().FirstOrDefault();
        if (projectMetadata != null)
        {
            try
            {
                var projectPath = projectMetadata.ProjectPath;
                using var projectCollection = new ProjectCollection();
                var project = projectCollection.LoadProject(projectPath);

                // .sqlprojx has a SqlTargetPath property, so try that first
                var targetPath = project.GetPropertyValue("SqlTargetPath");
                if (string.IsNullOrWhiteSpace(targetPath))
                {
                    targetPath = project.GetPropertyValue("TargetPath");
                }

                return targetPath;
            }
            catch (InvalidProjectFileException ex)
            {
                if (projectMetadata.ProjectPath.EndsWith(".sqlproj", StringComparison.OrdinalIgnoreCase))
                {
                    var dacpacPath = GetSqlprojDacpacPath(projectMetadata.ProjectPath);

                    if (dacpacPath != null)
                    {
                        return dacpacPath;
                    }
                }

                throw new InvalidOperationException($"Failed to load SQL Server Database project file {projectMetadata.ProjectPath}.", ex);
            }
        }

        var dacpacMetadata = Annotations.OfType<DacpacMetadataAnnotation>().FirstOrDefault();
        if (dacpacMetadata != null)
        {
            return dacpacMetadata.DacpacPath;
        }

        throw new InvalidOperationException($"Unable to locate SQL Server Database project package for resource {Name}.");
    }

    private string? GetSqlprojDacpacPath(string projectPath)
    {
        var directory = Path.GetDirectoryName(projectPath);

        if (directory == null)
        {
            return null;
        }

        var searchPath = Path.Combine(directory, "bin\\debug");

        var file = FindFile(searchPath);

        if (file != null)
        {
            return file;
        }

        searchPath = Path.Combine(directory, "bin\\release");

        return FindFile(searchPath);
    }

    private static string? FindFile(string searchPath)
    {
        if (!Directory.Exists(searchPath))
        {
            return null;
        }

        var files = Directory.GetFiles(searchPath, "*.dacpac", SearchOption.AllDirectories)
            .Where(f => !f.EndsWith("\\msdb.dacpac", StringComparison.OrdinalIgnoreCase)
                && !f.EndsWith("\\master.dacpac", StringComparison.OrdinalIgnoreCase))
            .ToList();

        if (files.Count == 1)
        {
            return files[0];
        }

        return null;
    }
}

```

jmezach:
Also added a redeploy command that looks like this:

Screenshot 2024-10-18 at 16 12 57

It is only available when the SQL project has finished deploying successfully at least once. When clicking redeploy it will go through the same flow again. Can be really useful I think during development.

ErikEJ:
Cool, love the redeploy, very useful. And the eventing makes scenarios like Data Api Builder so simple! (In particular when a Data Api Builder hosting component is also available)

https://github.com/JerryNixon/aspire-sqlserver/blob/main/202407-AspireDay/AspireDay.AppHost/Program.cs

ErikEJ:
I have just confirmed that project discovery also works with the new .sqlproj projects supported by Azure Data Studio / VS Code

jmezach:
Not entirely sure what you mean with that ;).

kzryzstof:
Is this a known problem that the package fails if the `.net SDK 8.0.403` is used?

However it seems to work properly with the `SDK 8.0.204`. This package is very useful :)

Here is the error that I got:

`Microsoft.Build.Exceptions.InvalidProjectFileException: Invalid static method invocation syntax: "[MSBuild]::SubstringByAsciiChars($(MSBuildProjectFile), 0, 8)". Method '[MSBuild]::SubstringByAsciiChars' not found. Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)). Check that all parameters are defined, are of the correct type, and are specified in the right order.  /usr/local/share/dotnet/sdk/8.0.403/Microsoft.Common.CurrentVersion.targets
   at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject(String errorSubCategoryResourceName, IElementLocation elementLocation, String resourceName, Object[] args)
   at Microsoft.Build.Shared.ProjectErrorUtilities.ThrowInvalidProject[T1,T2](IElementLocation elementLocation, String resourceName, T1 arg0, T2 arg1)
   at Microsoft.Build.Evaluation.Expander`2.Function`1.Execute(Object objectInstance, IPropertyProvider`1 properties, ExpanderOptions options, IElementLocation elementLocation)
   at Microsoft.Build.Evaluation.Expander`2.PropertyExpander`1.ExpandPropertyBody(String propertyBody, Object propertyValue, IPropertyProvider`1 properties, ExpanderOptions options, IElementLocation elementLocation, UsedUninitializedProperties usedUninitializedProperties, IFileSystem fileSystem)
   at Microsoft.Build.Evaluation.Expander`2.PropertyExpander`1.ExpandPropertiesLeaveTypedAndEscaped(String expression, IPropertyProvider`1 properties, ExpanderOptions options, IElementLocation elementLocation, UsedUninitializedProperties usedUninitializedProperties, IFileSystem fileSystem, LoggingContext loggingContext)
   at Microsoft.Build.Evaluation.Expander`2.PropertyExpander`1.ExpandPropertiesLeaveEscaped(String expression, IPropertyProvider`1 properties, ExpanderOptions options, IElementLocation elementLocation, UsedUninitializedProperties usedUninitializedProperties, IFileSystem fileSystem, LoggingContext loggingContext)
   at Microsoft.Build.Evaluation.Expander`2.ExpandIntoStringLeaveEscaped(String expression, ExpanderOptions options, IElementLocation elementLocation, LoggingContext loggingContext)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluatePropertyElement(ProjectPropertyElement propertyElement)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluatePropertyGroupElement(ProjectPropertyGroupElement propertyGroupElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluateImportElement(String directoryOfImportingFile, ProjectImportElement importElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluateImportElement(String directoryOfImportingFile, ProjectImportElement importElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluateImportElement(String directoryOfImportingFile, ProjectImportElement importElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluateImportElement(String directoryOfImportingFile, ProjectImportElement importElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluateImportElement(String directoryOfImportingFile, ProjectImportElement importElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.EvaluateImportElement(String directoryOfImportingFile, ProjectImportElement importElement)
   at Microsoft.Build.Evaluation.Evaluator`4.PerformDepthFirstPass(ProjectRootElement currentProjectOrImport)
   at Microsoft.Build.Evaluation.Evaluator`4.Evaluate()
   at Microsoft.Build.Evaluation.Evaluator`4.Evaluate(IEvaluatorData`4 data, Project project, ProjectRootElement root, ProjectLoadSettings loadSettings, Int32 maxNodeCount, PropertyDictionary`1 environmentProperties, ILoggingService loggingService, IItemFactory`2 itemFactory, IToolsetProvider toolsetProvider, IDirectoryCacheFactory directoryCacheFactory, ProjectRootElementCacheBase projectRootElementCache, BuildEventContext buildEventContext, ISdkResolverService sdkResolverService, Int32 submissionId, EvaluationContext evaluationContext, Boolean interactive)
   at Microsoft.Build.Evaluation.Project.ProjectImpl.Reevaluate(ILoggingService loggingServiceForEvaluation, ProjectLoadSettings loadSettings, EvaluationContext evaluationContext)
   at Microsoft.Build.Evaluation.Project.ProjectImpl.ReevaluateIfNecessary(ILoggingService loggingServiceForEvaluation, ProjectLoadSettings loadSettings, EvaluationContext evaluationContext)
   at Microsoft.Build.Evaluation.Project.ProjectImpl.ReevaluateIfNecessary(ILoggingService loggingServiceForEvaluation, EvaluationContext evaluationContext)
   at Microsoft.Build.Evaluation.Project.ProjectImpl.ReevaluateIfNecessary(EvaluationContext evaluationContext)
   at Microsoft.Build.Evaluation.Project.ProjectImpl.Initialize(IDictionary`2 globalProperties, String toolsVersion, String subToolsetVersion, ProjectLoadSettings loadSettings, EvaluationContext evaluationContext, Boolean interactive)
   at Microsoft.Build.Evaluation.Project..ctor(String projectFile, IDictionary`2 globalProperties, String toolsVersion, String subToolsetVersion, ProjectCollection projectCollection, ProjectLoadSettings loadSettings, EvaluationContext evaluationContext, IDirectoryCacheFactory directoryCacheFactory, Boolean interactive)
   at Microsoft.Build.Evaluation.Project..ctor(String projectFile, IDictionary`2 globalProperties, String toolsVersion, String subToolsetVersion, ProjectCollection projectCollection, ProjectLoadSettings loadSettings)
   at Microsoft.Build.Evaluation.Project..ctor(String projectFile, IDictionary`2 globalProperties, String toolsVersion, ProjectCollection projectCollection, ProjectLoadSettings loadSettings)
   at Microsoft.Build.Evaluation.Project..ctor(String projectFile, IDictionary`2 globalProperties, String toolsVersion, ProjectCollection projectCollection)
   at Microsoft.Build.Evaluation.Project..ctor(String projectFile, IDictionary`2 globalProperties, String toolsVersion)
   at Microsoft.Build.Evaluation.Project..ctor(String projectFile)
   at MSBuild.Sdk.SqlProj.Aspire.SqlProjectResource.GetDacpacPath() in /_/src/MSBuild.Sdk.SqlProj.Aspire/SqlProjectResource.cs:line 15
   at MSBuild.Sdk.SqlProj.Aspire.PublishSqlProjectLifecycleHook.AfterResourcesCreatedAsync(DistributedApplicationModel application, CancellationToken cancellationToken) in /_/src/MSBuild.Sdk.SqlProj.Aspire/PublishSqlProjectLifecycleHook.cs:line 26
   at Aspire.Hosting.Dcp.ApplicationExecutor.RunApplicationAsync(CancellationToken cancellationToken) in /_/src/Aspire.Hosting/Dcp/ApplicationExecutor.cs:line 141
   at Aspire.Hosting.Dcp.DcpHostService.StartAsync(CancellationToken cancellationToken) in /_/src/Aspire.Hosting/Dcp/DcpHostService.cs:line 70
   at Microsoft.Extensions.Hosting.Internal.Host.<StartAsync>b__15_1(IHostedService service, CancellationToken token)
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](IEnumerable`1 services, CancellationToken token, Boolean concurrent, Boolean abortOnFirstException, List`1 exceptions, Func`3 operation)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Aspire.Hosting.DistributedApplication.RunAsync(CancellationToken cancellationToken) in /_/src/Aspire.Hosting/DistributedApplication.cs:line 311
   at Program.<Main>$(String[] args) in`

ErikEJ:
@kzryzstof See this: https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/585#issuecomment-2250072052 - do you have a preview SDK installed? Try a global.json file.

kzryzstof:
@ErikEJ So I added the global.json file as suggested and choose the 8.0.403 (not a preview) SDK. It still failed with the same error :(

All my projects are running `.net8` (`8.2.2` for Aspire).


**Rider configuration**
Screenshot 2024-11-11 at 9 59 35 AM

**Global.json**
Screenshot 2024-11-11 at 9 59 57 AM

**Teh error**
Screenshot 2024-11-11 at 10 00 39 AM

jmezach:
We have decided that this integration is better served from the [CommunityToolkit for Aspire](https://github.com/CommunityToolkit/Aspire) rather than integrating it here. Thanks to everyone who has tried out this feature. We highly recommend migrating to [CommunityToolkit version](https://www.nuget.org/packages/CommunityToolkit.Aspire.Hosting.SqlDatabaseProjects/) of this integration.

PR #598, opened 2024-08-01
Report DacModelExceptions better, and continue processing all files

### LINKED ISSUE #595: Filename missing in the error message
When error occurs in project, the filename of "what file had error" would be important information to see in the error info.
`SQL71006 Only one statement is allowed per batch. A batch separator, such as 'GO', might be required between statements.`

image

**Background**
I'm using this with a database with around hundred tables, and merge constantly branches of different SQL schemas.
Here is an example, where things go wrong, because automatic merge doesn't understand to add a `GO` statement:
image
It would be very helpful to tell which of the modified 30+ files was the one that had the merge-error.

**Workaround**
Create old-fashion .sqlproj file next to this project and include the files there as well.
When you compile that, it will actually tell which file has the error.

ErikEJ:
fixes #595

### DIFF
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -72,6 +72,30 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             return outputMessageBuilder.ToString();
         }
 
+        public static string Format(this DacModelException exception, string fileName)
+        {
+            var stringBuilder = new StringBuilder();
+
+            foreach (var modelError in exception.Messages)
+            {
+                stringBuilder.Append(fileName);
+                stringBuilder.Append('(');
+                stringBuilder.Append("1");
+                stringBuilder.Append(',');
+                stringBuilder.Append("1");
+                stringBuilder.Append("):");
+                stringBuilder.Append(' ');
+                stringBuilder.Append("Error");
+                stringBuilder.Append(' ');
+                stringBuilder.Append(modelError.Prefix);
+                stringBuilder.Append(modelError.Number);
+                stringBuilder.Append(": ");
+                stringBuilder.Append(modelError.Message);
+            }
+
+            return stringBuilder.ToString();
+        }
+
         public static string GetPreDeploymentScript(this DacPackage package)
         {
             var stream = package.PreDeploymentScript;
--- a/src/DacpacTool/PackageBuilder.cs
+++ b/src/DacpacTool/PackageBuilder.cs
@@ -12,15 +12,21 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
 {
     public sealed class PackageBuilder : IDisposable
     {
+        private readonly IConsole _console;
         private bool? _modelValid;
 
         private List<int> _suppressedWarnings = new ();
         private Dictionary<string,List<int>> _suppressedFileWarnings = new Dictionary<string, List<int>>(StringComparer.InvariantCultureIgnoreCase);
 
+        public PackageBuilder(IConsole console)
+        {
+            _console = console ?? throw new ArgumentNullException(nameof(console));
+        }
+
         public void UsingVersion(SqlServerVersion version)
         {
             Model = new TSqlModel(version, Options);
-            Console.WriteLine($"Using SQL Server version {version}");
+            _console.WriteLine($"Using SQL Server version {version}");
         }
 
         public void AddReference(string referenceFile, string externalParts = null, bool suppressErrorsForMissingDependencies = false)
@@ -30,7 +36,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
 
             ValidateReference(referenceFile);
 
-            Console.WriteLine($"Adding reference to {referenceFile} with external parts {externalParts} and SuppressMissingDependenciesErrors {suppressErrorsForMissingDependencies}");
+            _console.WriteLine($"Adding reference to {referenceFile} with external parts {externalParts} and SuppressMissingDependenciesErrors {suppressErrorsForMissingDependencies}");
             Model.AddReference(referenceFile, externalParts, suppressErrorsForMissingDependencies);
         }
 
@@ -76,7 +82,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 return true;
             }
 
-            Console.WriteLine($"Adding {inputFile.FullName} to the model");
+            _console.WriteLine($"Adding {inputFile.FullName} to the model");
 
             try
             {
@@ -86,7 +92,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             }
             catch (DacModelException dex)
             {
-                Console.WriteLine(dex.Format(inputFile.Name));
+                _console.WriteLine(dex.Format(inputFile.Name));
                 return false;
             }
         }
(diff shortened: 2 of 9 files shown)


ErikEJ:
image

ErikEJ:
```dos
Build started at 07:30...
1>------ Build started: Project: TestProjectWithErrors, Configuration: Debug Any CPU ------
1>Using target framework net8.0 to run DacpacTool
1>Using package name TestProjectWithErrors and version 1.0.0
1>Using SQL Server version Sql150
1>Adding C:\Code\Github\MSBuild.Sdk.SqlProj\test\TestProjectWithErrors\Tables\MyTable.sql to the model
1>MyTable.sql(1,1): error SQL71006: Only one statement is allowed per batch. A batch separator, such as 'GO', might be required between statements.
1>Adding C:\Code\Github\MSBuild.Sdk.SqlProj\test\TestProjectWithErrors\Tables\MyTable2.sql to the model
1>MyTable2.sql(1,1): error SQL71006: Only one statement is allowed per batch. A batch separator, such as 'GO', might be required between statements.
1>C:\Code\Github\MSBuild.Sdk.SqlProj\src\MSBuild.Sdk.SqlProj\Sdk\Sdk.targets(243,5): error MSB3073: The command "dotnet "C:\Code\Github\MSBuild.Sdk.SqlProj\src\MSBuild.Sdk.SqlProj\Sdk\../tools/net8.0/DacpacTool.dll" build -o "obj\Debug\netstandard2.0\TestProjectWithErrors.dacpac" -n "TestProjectWithErrors" -v "1.0.0" -sv Sql150 -i "obj\Debug\netstandard2.0\TestProjectWithErrors.InputFiles.txt"    -dp IncludeCompositeObjects=true           " exited with code 1.
1>Done building project "TestProjectWithErrors.csproj" -- FAILED.
========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========
========== Build completed at 07:30 and took 01,579 seconds ==========

```

test/DacpacTool.Tests/PackageBuilderTests.cs
```diff
@@ -47,12 +46,27 @@ public void AddInputFile_FileExists()
...
+        {
+            // Arrange
+            var packageBuilder = new PackageBuilder();
+            packageBuilder.UsingVersion(SqlServerVersion.Sql150);
+
+            // Act
+            var result = packageBuilder.AddInputFile(new FileInfo("../../../../TestProjectWithExceptions/Tables/MyTable.sql"));
+
+            // Assert
+            result.ShouldBe(false);
```
jeffrosenberg, line 67: Is it possible to test the exceptions that are thrown, as well as the result of `AddInputFile`?

src/DacpacTool/Extensions.cs
```diff
@@ -72,6 +72,30 @@ public static string Format(this BatchParserExecutionErrorEventArgs args, string
...
         }
 
+        public static string Format(this DacModelException exception, string fileName)
+        {
+            var stringBuilder = new StringBuilder();
+
+            foreach (var modelError in exception.Messages)
+            {
+                stringBuilder.Append(fileName);
+                stringBuilder.Append('(');
```
jeffrosenberg, line 82: Is the `(1,1)` here just to fit with the formatting of other errors? Is it required to be in that format? If it's not required, I'd personally skip it, since in my opinion it may cause people to look in the wrong place for an error.

jeffrosenberg (approved):

test/DacpacTool.Tests/PackageBuilderTests.cs
ErikEJ: Do you mean assert what is written to the console?

src/DacpacTool/Extensions.cs
ErikEJ: This is to help errors display properly in the VS error list. The convention is to use 1,1 is the exact file location is not available.

test/DacpacTool.Tests/PackageBuilderTests.cs
ErikEJ: @jeffrosenberg Added assertion of console output

src/DacpacTool/PackageBuilder.cs
```diff
@@ -73,11 +73,22 @@ public void AddInputFile(FileInfo inputFile)
...
+
+            try
+            {
+                TSqlObjectOptions sqlObjectOptions = new TSqlObjectOptions();
+                Model.AddOrUpdateObjects(File.ReadAllText(inputFile.FullName), inputFile.FullName, new TSqlObjectOptions());
+                return true;
+            }
+            catch (DacModelException dex)
+            {
+                Console.WriteLine(dex.Format(inputFile.Name));
```
jmezach: This might be beyond the scope of this PR, but I'm now wondering why this isn't using the `IConsole` abstraction instead of directly writing to the `Console`.

src/DacpacTool/Extensions.cs
ErikEJ: Confirmed the 1,1 convention

test/DacpacTool.Tests/PackageBuilderTests.cs
jmezach: ;) This is kind of why we have the `IConsole` abstraction. Not entirely sure why the `PackageBuilder` class didn't get this treatment. My guess is that this is the first thing I wrote so hadn't thought about it at the time. Might be worth changing now though.

test/TestProjectWithExceptions/Tables/MyTable.sql
jmezach, line 1: Maybe it is just me, but I'm not seeing any different between `MyTable.sql` and `MyTable2.sql`. Is the expectation here that one successfully adds to the model and the other one does not? If so we might want to add a comment explaining that or something.

ErikEJ: None of the scrips gets add, they both throw a DacModelException

ErikEJ: @jmezach Should I change anything?

jmezach: Perhaps add a comment to the files then, to indicate that they are designed to fail?

ErikEJ:
Phew - that was an "interesting" breaking SQL container change 😄

ErikEJ:
@jmezach All lights are green now 😄

jmezach:
Looks like I can't approve this from my phone. Will do so when I get back home in about half an hour ;)

jmezach:
> Phew - that was an "interesting" breaking SQL container change 😄

Looks like we weren't the only ones hit by this breaking change: https://github.com/microsoft/mssql-docker/issues/892

jmezach (approved):
Let's get this merged, nice fix I think :)

ErikEJ:
@jmezach Agree, this is a good quality improvement

jmezach:
Should we release this as 2.8.1?

ErikEJ:
I have a plan for a PR to run some .NET analyzers against the DacpacTool codebase (our code, not the modules) - maybe after that?

jmezach:
Sure thing :)

src/DacpacTool/Extensions.cs
jeffrosenberg: Thanks for clarifying, I didn't realize that was a convention 👍

PR #601, opened 2024-08-02
Code cleanup from temporary enabling .NET analyzers

ErikEJ:
Also: removed .NET 7 support and updated readme to 2.8.1

```xml
<AnalysisMode>all</AnalysisMode>
    <EnableNETAnalyzers>true</EnableNETAnalyzers>
    <AnalysisLevel>latest</AnalysisLevel>
    <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,16 +26,11 @@ jobs:
         fetch-depth: 0 # avoid shallow clone so nbgv can do its work.
         submodules: 'true' # need to check out the SqlTools submodule to build successfully.
 
-    # Install .NET SDK's for 6.0 and 7.0
+    # Install .NET SDK's for 6.0 and 8.0
     - name: Setup dotnet 6.0
       uses: actions/setup-dotnet@v4
       with:
         dotnet-version: '6.0.x'
-
-    - name: Setup dotnet 7.0
-      uses: actions/setup-dotnet@v4
-      with:
-        dotnet-version: '7.0.x'
 
     - name: Setup dotnet 8.0
       uses: actions/setup-dotnet@v4
@@ -112,7 +107,7 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-22.04", "macos-13", "windows-2022" ]
-        dotnet: [ '6.0.x', '7.0.x', '8.0.x' ]
+        dotnet: [ '6.0.x', '8.0.x' ]
       fail-fast: false
 
     steps:
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
     <LangVersion>9.0</LangVersion>
   </PropertyGroup>
 
--- a/src/DacpacTool/DeployOptions.cs
+++ b/src/DacpacTool/DeployOptions.cs
@@ -13,6 +13,6 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         public string[] Property { get; set; }
         public string[] SqlCmdVar { get; set; }
         public bool RunScriptsFromReferences { get; set; }
-        public bool Encrypt { get; set; } = false;
+        public bool Encrypt { get; set; }
     }
 }
--- a/src/DacpacTool/PackageBuilder.cs
+++ b/src/DacpacTool/PackageBuilder.cs
@@ -11,7 +11,7 @@ using Microsoft.SqlServer.Dac.Model;
 
 namespace MSBuild.Sdk.SqlProj.DacpacTool
 {
-    internal sealed class PackageBuilder : IDisposable
+    public sealed class PackageBuilder : IDisposable
     {
         private readonly IConsole _console;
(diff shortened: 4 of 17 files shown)


src/DacpacTool/DacpacTool.csproj
jmezach, line 1: I think I made the same change in my .NET Aspire branch as well, but I think we can merge this first and then I'll do a rebase and resolve the conflicts.

src/DacpacTool/DeployOptions.cs
jmezach, line 1: Is this going to be a breaking change for our users?

src/DacpacTool/DacpacTool.csproj
ErikEJ: Cool, I thought it was partly already in master. I should update the pipeline as well then.

src/DacpacTool/DeployOptions.cs
ErikEJ: No, the default value is always false

jmezach:
Is there any reason not to keep these analyzers on?

ErikEJ:
> Is there any reason not to keep these analyzers on?

@jmezach Yes, they produce a lot of warnings from the module code (that we cannot modify)

jmezach:
You mean the `sqltoolsservice` submodule? Yeah, I've been wondering if we can somehow get rid of that dependency as we haven't been able to update it for quite a while now.

ErikEJ:
Getting rid of it would be great - looks like there is also ways to exclude files and/or  folder from analysis https://github.com/dotnet/roslyn/issues/3705

ErikEJ:
We could maybe log an improvement issue to get this done?

jmezach:
Sounds good to me

ErikEJ:
Logged https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/602

jmezach (approved):

.github/workflows/main.yml
```diff
@@ -112,7 +107,7 @@ jobs:
     strategy:
       matrix:
         os: [ "ubuntu-22.04", "macos-13", "windows-2022" ]
-        dotnet: [ '6.0.x', '7.0.x', '8.0.x' ]
```
jeffrosenberg, line 115: Is there any harm in leaving `7.0.x`? Feels to me like it might be good to include all supported versions. Or maybe we want to remove `6.0.x` instead, as it will be going out of support shortly.

src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj
```diff
@@ -31,7 +31,7 @@
 
   <Target Name="IncludeDacpacTool" AfterTargets="Build">
     <PropertyGroup>
-      <_DacpacToolSupportedTfms>net6.0;net7.0;net8.0</_DacpacToolSupportedTfms>
```
jeffrosenberg, line 34: Similar question about `net7.0` support, shouldn't we be supporting this?

src/DacpacTool/PackageBuilder.cs
```diff
@@ -40,7 +41,7 @@ public void AddReference(string referenceFile, string externalParts = null, bool
             Model.AddReference(referenceFile, externalParts, suppressErrorsForMissingDependencies);
         }
 
-        private void ValidateReference(string referenceFile)
+        private static void ValidateReference(string referenceFile)
```
jeffrosenberg, line 44: Can you explain why we're making this static? From what I can tell, it looks like we use this as an instance (i.e. `using var packageBuilder = new PackageBuilder(new ActualConsole());`). I did a quick search and don't see any calls to static methods.

src/DacpacTool/PackageDeployer.cs
```diff
@@ -125,9 +125,9 @@ public void RunPostDeploymentScriptFromReferences(FileInfo dacpacPackage, string
         private void RunDeploymentScriptFromReferences(FileInfo dacpacPackage, string targetDatabaseName, bool isPreDeploy)
         {
             using var model = new TSqlModel(dacpacPackage.FullName, DacSchemaModelStorageType.Memory);
-            var references = model.GetReferencedDacPackages();
+            var references = model.GetReferencedDacPackages().ToList();
```
jeffrosenberg, line 128: I don't necessarily have a problem with this, but I'm not certain why using `ToList()` and `Count` is better than using `Any()`

jeffrosenberg (approved):
A few minor comments... I'm going to go ahead and approve, since I probably won't be able to look back at this PR for 12 hours or so

.github/workflows/main.yml
ErikEJ: 7.0 is out of support as of May this year

src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj
ErikEJ: 7.0 is no longer supported

src/DacpacTool/PackageDeployer.cs
ErikEJ: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1851 - I will probably revisit in #602

.github/workflows/main.yml
jeffrosenberg: Whoops, guess I need to get with the times!

PR #632, opened 2024-10-03
Add a version check and warn if not using latest

### LINKED ISSUE #629: Investigate a way for the SDK to test for latest version
Maybe something like this coudl be used:

https://github.com/ErikEJ/EFCorePowerTools/blob/master/src/Core/efcpt.8/Services/PackageService.cs#L22

ErikEJ:
fixes #629

### DIFF
--- a/src/DacpacTool/Program.cs
+++ b/src/DacpacTool/Program.cs
@@ -45,9 +45,9 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 new Option<bool>(new string[] { "--debug" }, "Waits for a debugger to attach")
 #endif
             };
-            buildCommand.Handler = CommandHandler.Create<BuildOptions>(async (x) =>
+            buildCommand.Handler = CommandHandler.Create<BuildOptions>(async (buildOptions) =>
             {
-                await BuildDacpac(x).ConfigureAwait(false);
+                await BuildDacpac(buildOptions).ConfigureAwait(false);
             });
 
             var collectIncludesCommand = new Command("collect-includes")
--- a/src/DacpacTool/Program.cs
+++ b/src/DacpacTool/Program.cs
@@ -45,7 +45,10 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 new Option<bool>(new string[] { "--debug" }, "Waits for a debugger to attach")
 #endif
             };
-            buildCommand.Handler = CommandHandler.Create<BuildOptions>(BuildDacpac);
+            buildCommand.Handler = CommandHandler.Create<BuildOptions>(async (x) =>
+            {
+                await BuildDacpac(x).ConfigureAwait(false);
+            });
 
             var collectIncludesCommand = new Command("collect-includes")
             {
@@ -79,18 +82,18 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             rootCommand.Description = "Command line tool for generating a SQL Server Data-Tier Application Framework package (dacpac)";
 
             var processed = rootCommand.Parse(args);
-#pragma warning disable CA2007 // Consider calling ConfigureAwait on the awaited task
-            return await processed.InvokeAsync();
-#pragma warning restore CA2007 // Consider calling ConfigureAwait on the awaited task
+            return await processed.InvokeAsync().ConfigureAwait(false);
         }
 
-        private static int BuildDacpac(BuildOptions options)
+        private static async Task<int> BuildDacpac(BuildOptions options)
         {
             // Wait for a debugger to attach
             WaitForDebuggerToAttach(options);
 
             using var packageBuilder = new PackageBuilder(new ActualConsole());
             var versionChecker = new VersionChecker(new ActualConsole(), new VersionProvider());
+
+            await versionChecker.CheckForPackageUpdateAsync().ConfigureAwait(false);
 
             // Set metadata for the package
             packageBuilder.SetMetadata(options.Name, options.Version);
@@ -147,6 +150,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             {
                 if (options.InputFile.Exists)
                 {
+#pragma warning disable CA1849 // Call async methods when in an async method - must wait for .NET 6 to be removed
                     foreach (var line in File.ReadLines(options.InputFile.FullName))
                     {
                         FileInfo inputFile = new FileInfo(line); // Validation occurs in AddInputFile
@@ -155,6 +159,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                             modelExceptions = true;
                         }
                     }
+#pragma warning restore CA1849 // Call async methods when in an async method
                 }
                 else
                 {
@@ -174,6 +179,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             {
                 if (options.SuppressWarningsListFile.Exists)
                 {
+#pragma warning disable CA1849 // Call async methods when in an async method
                     foreach (var line in File.ReadLines(options.SuppressWarningsListFile.FullName))
                     {
                         //Checks if there are suppression warnings list
@@ -183,6 +189,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                         FileInfo inputFile = new FileInfo(parts[0]); // Validation occurs in AddInputFile
                         packageBuilder.AddFileWarningsToSuppress(inputFile, warningList);
                     }
+#pragma warning restore CA1849 // Call async methods when in an async method
                 }
             }
 
@@ -211,8 +218,6 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
 
                 analyzer.Analyze(packageBuilder.Model, options.Output, options.CodeAnalysisAssemblies ?? Array.Empty<FileInfo>());
             }
-
-            versionChecker.CheckForPackageUpdateAsync().GetAwaiter().GetResult();
 
             return 0;
         }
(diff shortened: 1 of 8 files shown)


ErikEJ:
image

src/DacpacTool/Program.cs
```diff
@@ -210,6 +212,8 @@ private static int BuildDacpac(BuildOptions options)
                 analyzer.Analyze(packageBuilder.Model, options.Output, options.CodeAnalysisAssemblies ?? Array.Empty<FileInfo>());
             }
 
+            versionChecker.CheckForPackageUpdateAsync().GetAwaiter().GetResult();
```
jmezach, line 215: Can't we just use `await` here?

src/DacpacTool/VersionChecker.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public async Task CheckForPackageUpdateAsync()
+        {
+            try
+            {
+                var logger = new NullLogger();
+
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+
+                using var cache = new SourceCacheContext();
```
jmezach, line 31: Hmm, how does this cache work? I'm a bit worried that this might end-up going to NuGet.org every time we run the `DacpacTool`, which is essentially on every build. That has the potential of significantly slowing down build times which is in the developers inner loop. Could we perhaps cache something ourselves or is there some prior art that we can look at?

src/DacpacTool/VersionProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool
+{
+    public class VersionProvider : IVersionProvider
+    {
+        public NuGetVersion CurrentPackageVersion()
+        {
+            return new NuGetVersion(
+                Assembly
+                .GetEntryAssembly()!
```
jmezach, line 12: Wouldn't this be the same as `typeof(VersionProvider).Assembly`?

src/DacpacTool/Program.cs
jmezach, line 215: We are doing this as the last step in the process, which makes sense from a performance perspective, but if an exception occurs somewhere before this you don't get this warning, while upgrading to the latest version has the potential of resolving the exception so do we might want to do this at the start instead?

ErikEJ: Have to make the method async then

src/DacpacTool/VersionChecker.cs
ErikEJ: I was thinking using the latest release info from GitHub might be more lightweight, maybe combined with some caching?

src/DacpacTool/VersionProvider.cs
ErikEJ: It would

src/DacpacTool/VersionChecker.cs
ErikEJ: https://api.github.com/repos/rr-wfm/MSBuild.Sdk.SqlProj/releases/latest

ErikEJ: No very well documented, but as I read it the cache has a 30 minute life time

ErikEJ: I also set a 5 second timeout (could maybe be made shorter) for the version check.

ErikEJ: @jmezach I have refactored the check to use the GitHub API and only call it once per day

ErikEJ:
@jmezach Thoughts after the updated approach?

src/DacpacTool/Program.cs
```diff
@@ -45,7 +45,10 @@ static async Task<int> Main(string[] args)
                 new Option<bool>(new string[] { "--debug" }, "Waits for a debugger to attach")
 #endif
             };
-            buildCommand.Handler = CommandHandler.Create<BuildOptions>(BuildDacpac);
+            buildCommand.Handler = CommandHandler.Create<BuildOptions>(async (x) =>
```
jmezach, line 48: I believe `x` are the `BuildOptions`? Maybe rename it to `options` to be clear?

src/DacpacTool/VersionChecker.cs
```diff
@@ -0,0 +1,85 @@
...
+                NuGetVersion latestVersion = null;
+
+                if (File.Exists(cacheFile))
+                {
+                    var cache = await File.ReadAllTextAsync(cacheFile, cts.Token).ConfigureAwait(false);
+                    latestVersion = NuGetVersion.Parse(cache);
+                }
+                else
+                {
+                    using var httpClient = new System.Net.Http.HttpClient
```
jmezach, line 42: Let's just add a using for `System.Net.Http`

src/DacpacTool/VersionChecker.cs
```diff
@@ -0,0 +1,85 @@
...
+                    {
+                        DefaultRequestHeaders = { { "User-Agent", "MSBuild.Sdk.SqlProj" } },
+                        Timeout = timeOut,
+                    };
+
+                    var response = await httpClient.GetFromJsonAsync<Release>("https://api.github.com/repos/rr-wfm/MSBuild.Sdk.SqlProj/releases/latest").ConfigureAwait(false);
+                    if (response is null || response.tag_name is null || response.draft || response.prerelease)
```
jmezach, line 49: Hmm, if we publish a prerelease or draft release then we're not doing the version check at all? Does that make sense?

src/DacpacTool/VersionChecker.cs
```diff
@@ -0,0 +1,85 @@
...
+                }
+
+                if (latestVersion is null)
+                {
+                    return;
+                }
+
+                if (latestVersion > _versionProvider.CurrentPackageVersion())
+                {
+                    _console.WriteLine($"DacpacTool warning SQLPROJ0002: Your are not using the latest version of this SDK, please update to get the latest bug fixes, features and support. Modify your project file: '<Project Sdk=\"MSBuild.Sdk.SqlProj/{latestVersion}\">')");
```
jmezach, line 66: You (not your)

jmezach (approved):
Just some small comments

src/DacpacTool/VersionChecker.cs
ErikEJ: I think that makes sense?

ErikEJ: A prerelease or draft would be something you would explicitly opt in to use, I think

ErikEJ: I have changed to always warn

src/DacpacTool/VersionChecker.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public async Task CheckForPackageUpdateAsync()
+        {
+            try
+            {
+                var timeout = TimeSpan.FromSeconds(2);
+
+                using var cts = new CancellationTokenSource(timeout);
+
+                var cacheFile = Path.Join(Path.GetTempPath(), "MSBuild.Sdk.SqlProj.tag-" + DateTime.Now.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture) + ".txt");
```
jmezach, line 32: Now that I think of it, this is going to create a new file each day (you are using MSBuild.Sdk.SqlProj). It is probably relatively small, but we don't actively clean up these files anywhere, so over time this can accumulate a bit. Perhaps we just want a single file and use the last write time to determine if we need to refetch it?

src/DacpacTool/VersionChecker.cs
```diff
@@ -0,0 +1,84 @@
...
+            catch (Exception)
+            {
+                // Ignore
+            }
+#pragma warning restore CA1031
+        }
+    }
+
+
+    internal class Release
```
jmezach, line 80: Are we going to do anything about the warnings on this? Perhaps making this class private would be enough?

ErikEJ: I will address the warnings

src/DacpacTool/VersionChecker.cs
ErikEJ: Good idea!

ErikEJ:
@jmezach Updated

jmezach:
Looking good now. Perhaps @jeffrosenberg can have a look as well and then we can merge this.

jeffrosenberg (approved):
LGTM. Just a heads-up, I'm in between two vacations at the moment. I'm going to look through all outstanding PRs, but then won't be available for a week. Feel free to merge PRs without my review during that time if you'd like!

ErikEJ:
@jmezach Should we push a 2.9.2 release - or maybe just wait for 3.0 (preview) ?

jmezach:
I think this can wait for 3.0

ErikEJ:
I just added a 3.1.0 GitHub release and see this now:

image

jmezach:
Nice, seems like this is working now :).

ErikEJ:
@jmezach There is nothing in the docs about how to disable our warnings - should we expect that our users know how to do that?

jmezach:
I'd say this is pretty well documented, but I guess it could be worth pointing out?

medianick:
Indeed, this did break our zero-warnings-allowed builds this morning, unexpectedly. I had to dig a little and found this PR, so we prioritized the upgrade. I do think this would be good to be able to opt out of, or to be able to configure whether it's a build warning or just an informational message.

ErikEJ:
Just add 

`<NoWarn>SQLPROJ0002</NoWarn>`

to your project file (or upgrade) :smile:

ErikEJ:
@medianick Or probably better:

<NoWarn>$(NoWarn);SQLPROJ0002</NoWarn>

PR #633, opened 2024-10-04
attempt to fix #615

### LINKED ISSUE #615: Version 2.8.1 - master.dacpac reference is not found
Version: 2.8.1
Deployment on Docker image mcr.microsoft.com/mssql/server:2019-latest
Sqlpackage version tested with 150.* and the latest 162.3.566.1

sqlpackage returns error:

```
Initializing deployment (Start) 2024-08-30 12:41:29 *** No file was supplied for reference master.dacpac; deployment might fail. When package was created, the original referenced file was located C:\Users\username\.nuget\packages\microsoft.sqlserver.dacpacs.master\150.1.1/tools/master.dacpac. 2024-08-30 12:41:29 Initializing deployment (Failed)
```

Notes:
* master.dacpac is in the same directory as dacpac of database I want to deploy;
* master is referenced:
```xml
<Project Sdk="MSBuild.Sdk.SqlProj/2.8.1">
     ...
    <ItemGroup>
        <PackageReference Include="Microsoft.SqlServer.Dacpacs.Master" DacpacName="master" DatabaseVariableLiteralValue="master" Version="150.1.1" />
    </ItemGroup>
</Project>
```
* issue gone once ```<Project Sdk="MSBuild.Sdk.SqlProj/2.8.1">``` was changed to ```<Project Sdk="MSBuild.Sdk.SqlProj/2.7.2">```

ErikEJ:


src/DacpacTool/Extensions.cs
```diff
@@ -154,6 +154,8 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
                 }
             }
 
+            File.Copy(referencePath, Path.Combine(Path.GetDirectoryName(typeof(PackageBuilder).Assembly.Location), Path.GetFileName(referencePath)), true);
```
jmezach, line 157: I'm wondering if rather than copying the referenced `.dacpac` to the location of the tool we should instead copy it next to the output `.dacpac` we are building (which I believe we are already doing) and execute the tool with the working directory set to the output folder?

ErikEJ: Yeah, the test failure inidcates that it is a better solution. Problem is that in many of the tests we do not even have an output .dacpac...

jmezach: I think we can solve this within the MSBuild part and that the tool doesn't necessarily have to worry about it.

jmezach: The `<Exec>` task we are using has a `WorkingDirectory` property that we are currently not setting, meaning it will run the command in the folder containing the `.csproj`. I think we can change that to be the output folder.

ErikEJ: Feel free to update this PR !

jmezach: Did some experimenting, but now the problem becomes that the set of input files are relative to the project folder so now it is failing on adding those. Might need to think about this some more before we can fix this.

PR #656, opened 2024-11-12
SqlCmdVariable DefaultValue should not be written to .dacpac as per docs

### LINKED ISSUE #448: Defining SQLCMD variables in the project prevents them from being set via environment variables
It seems to me that defining the `<SqlCmdVariable> ... </SqlCmdVariable>` is actually worse than not defining it.

I am building a .dacpac with this SDK with a Post-Deploy script that uses variables like this:
```sql
IF (1 = ISNULL($(SQLCMD_DEPLOY_CONFIGS), 0))
BEGIN
# ...
END
```
I am **not** defining  `<SqlCmdVariable> ... </SqlCmdVariable>` anywhere.
1. Running `sqlpackage.exe /a:Publish ...` in a container where the SQLCMD_DEPLOY_CONFIGS environment variable is defined results in the value from the environment variable being replaced in the script.
2. Running `sqlpackage.exe /a:Publish ... /v:SQLCMD_DEPLOY_CONFIGS=0` also works, although it prints _"The following SqlCmd variables are not defined in the target scripts: SQLCMD_DEPLOY_CONFIGS."_

**BUT**

If I go and now define:
```xml
<SqlCmdVariable Include="SQLCMD_DEPLOY_CONFIGS">
		<!--Defines whether configuration scripts should be deployed: 0 - No, 1 - Yes-->
	<DefaultValue>$(SQLCMD_DEPLOY_CONFIGS)</DefaultValue>
	<Value>$(SQLCMD_DEPLOY_CONFIGS)</Value>
</SqlCmdVariable>
```

then the approach with the environment variable (1) **no longer works**. So to me it seems that it is worse to define the sqlcmd vars?

ErikEJ:
https://learn.microsoft.com/sql/tools/sql-database-projects/concepts/sqlcmd-variables

fixes #448

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ Especially when using pre- and post-deployment scripts, but also in other scenar
 </Project>
 ```
 
-> Note: The value of `Value` is checked first and if it found to be empty, we'll fall back to `DefaultValue`.
+> Note: With version 3.0.0 of the SDK, the `DefaultValue` is not applied to the build output, in line with the standard `.sqlproj` behaviour.
 
 ## Package references
 `MSBuild.Sdk.SqlProj` supports referencing NuGet packages that contain `.dacpac` packages. These can be referenced by using the `PackageReference` format familiar to .NET developers. They can also be installed through the NuGet Package Manager in Visual Studio.
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -260,18 +260,11 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             {
                 var varWithValue = variable.Split('=', 2);
                 var variableName = varWithValue[0];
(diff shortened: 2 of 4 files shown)


ErikEJ:
@jmezach Any idea why the build failed?

jmezach:
Yeah, it looks like we're being hit again by having a submodule dependency ;).

jmezach (approved):
LGTM. One thing I'm wondering whether we now need to pass the values of SqlCmd variables to the DacpacTool from the SDK at all since we're not doing anything with it anyway?

ErikEJ:
Correct, but this change just avoids any SDK changes

ErikEJ:
Maybe the docs need adjusted as well, I will check

ErikEJ:
@jmezach Updated the readme

jeffrosenberg (approved):

PR #661, opened 2024-11-14
[Docs] Show how to extract an existing database

ErikEJ:


jmezach:
I agree that this is at least related, but I was wondering if there's a doc on SqlPackage that we could point at for more details as well. We don't own that tool, so if things change we might not know about it ;)

ErikEJ:
@jmezach I think this is quite relevant, I often bootstrap a legacy .sqlproj by importing an existing database schema, and we do not have a VS gesture for that.

The docs from sqlpackage are quite poor in this area, the main docs are here: https://learn.microsoft.com/sql/tools/sqlpackage/sqlpackage-extract

jmezach:
I agree that we should add this. Let's just add the link to the doc for those who want to dig deeper :).

ErikEJ:
@jmezach Added a docs link

README.md
```diff
@@ -81,6 +81,19 @@ If you are using Visual Studio, to make files excluded from the package appear i
 
 Wildcards are supported for all nodes (`Content`, `None`, etc.). For example, `<None Include="Directory\**" />`.
 
+
+## Use an existing database
+
+If you want to initialize your project with objects from an existing database, you can do so with the following command:
+
+```
+sqlpackage /a:Extract /p:ExtractTarget=Flat /scs:"data source=.\SQLEXPRESS;initial catalog=Chinook;Trusted_Connection=true;encrypt=False" /tf:.\Tables
```
jeffrosenberg, line 90: I think it would be better to show placeholders here, instead of example values. This could then be followed by an example. I've seen too many people try to copy-paste examples like this and then be surprised when it doesn't work. I'd also recommend long parameters instead of short, since they're clearer.

```suggestion
sqlpackage /Action:Extract /Properties:ExtractTarget=Flat /SourceConnectionString:"<connection_string>" /TargetFile:<target_dacpac_file>
```

You could then add your existing command below, as long as it's clear that it's an example.

ErikEJ:
@jeffrosenberg Updated

jeffrosenberg (approved):
Thanks for the update, LGTM!

PR #676, opened 2024-12-29
Remove sqltoolsservice submodule

### LINKED ISSUE #675: Remove Dependency on sqltoolsservice submodule

ErikEJ:
fixes #675
parts of #674 (maybe all?)


test/DacpacTool.Tests/DacpacTool.Tests.csproj
```diff
@@ -10,13 +10,20 @@
     <TestTfmsInParallel>false</TestTfmsInParallel>
   </PropertyGroup>
 
+  <ItemGroup>
```
jmezach, line 13: Are these changes necessary?

src/DacpacTool/PackageDeployer.cs
```diff
@@ -182,13 +182,19 @@ private void RunDeploymentScriptFromReferences(FileInfo dacpacPackage, string ta
             }
         }
 
-        private void AddSqlCmdVariables(ScriptExecutionArgs args, string targetDatabaseName)
+        private Dictionary<string, string> AddSqlCmdVariables(string targetDatabaseName)
```
jmezach, line 185: Is this necessary? I was assuming this was done because `ScriptExecutionArgs` isn't available now, but we're still instantiating it now?

ErikEJ: Not sure what you mean - this method is used in the constructor above. ".Variables" is now internal...

src/DacpacTool/Microsoft.SqlTools.ManagedBatchParser.dll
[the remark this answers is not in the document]
ErikEJ: Yes, that is the controversial part of this 😢 - we could add an action to watch for new sqltoolsservice releases like we do for our own new releases on GitHub?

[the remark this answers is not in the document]
jeffrosenberg: Yeah, I would probably recommend using GitHub's [list releases API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) API to check whether we're behind, and how much. Potentially we could set a version number as a variable, and use an action to update it? 

If you'd be willing to merge this and create a new issue for keeping this dependency up to date, I could work on it in a week or two.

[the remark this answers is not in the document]
ErikEJ: Yes, that was my suggestion, we already use that API to check for updates to "ourselves" https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/src%2FDacpacTool%2FVersionChecker.cs

src/DacpacTool/PackageDeployer.cs
ErikEJ: @jmezach  Ping?

jmezach: Previously we created an instance of `ScriptExecutionArgs` and then passed that to `AddSqlCmdVariables`. Now we are calling `AddSqlCmdVariables` first and passing the result of that to the constructor of `ScriptExecutionArgs`. I'm just trying to understand why we made that change ;)

ErikEJ: @jmezach The ScriptExecutionArgs constructor changed, so the only way to set variables is via the constructor, as the Variables property on ScriptExecutionArgs no longer exists.

jmezach: Okay, yeah, that makes sense

src/DacpacTool/Microsoft.SqlTools.ManagedBatchParser.dll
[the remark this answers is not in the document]
ErikEJ: @jmezach Thoughts?

jeffrosenberg (approved):

src/DacpacTool/Microsoft.SqlTools.ManagedBatchParser.dll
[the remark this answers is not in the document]
jeffrosenberg: Yup, I agree too. I'd be fine with either a warning or an automated update to our dependency.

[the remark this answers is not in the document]
ErikEJ: @jmezach Looks like we are in agreement

ErikEJ:
@jmezach Are your concerns adressed?

jmezach (approved):
Oh yeah, sorry. This LGTM :).

ErikEJ:
@jmezach I know it is not ideal, but at least slightly better than the current submodule trouble

PR #700, opened 2025-01-29
Correct fenced-in code blocks

f3-epicor:
A couple of the code blocks have improper termination of their fenced-off markdown, and (in one case) were preventing the _Pre- and post deployment scripts_ subheading from rendering.  Clean up the code blocks so they (and markup that follows them) are rendered properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,8 @@ You can suppress warnings for a specific file by adding `SuppressTSqlWarnings` f
         </Content>
     </ItemGroup>
 </Project>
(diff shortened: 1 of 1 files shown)


jeffrosenberg (approved):
Thanks for catching this, we appreciate the help!

@jmezach @ErikEJ can one of you review, please? Do we want to log an issue to track this against, or are we okay as-is?

ErikEJ (approved):

PR #702, opened 2025-01-30
stop passing sqlcmd default values when launching dacpactool

### LINKED ISSUE #701: Issues with SQLCMD special characters going from 2.6.0 -> 2.6.1 / 3.0.0
Hi, upon upgrading from 2.6.0 to 3.0.0 we started having problems deploying (specifically our pre/post-deploy scripts), and saw this change in 2.6.1:  https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/459/files

We confirmed that we have problems with 2.6.1 as well, but they manifested differently.  As such we started focusing on 3.0.0 for a minimal repro.  We've confirmed that the issue is with SQLCMD variables with special characters.

We tried to develop a minimal repro, however we kept finding more and more problematic characters so it has become increasingly complex.  I assume it's some same root cause.  I'm attaching it here ([MsBuildSqlTest.zip](https://github.com/user-attachments/files/18594722/MsBuildSqlTest.zip)) with instructions below.

**In the meantime, is there any known way in 3.0.0 to use SQLCMD values with special characters?  (`&`, `<`, ` ` (space), etc.)?**  
Previously we simply had to wrap the value in quotes.  Space seems to work (TestValue3), but the rest seem to hopelessly break the command line.

Instructions:
1. have a target SQL server available
2. have an environment with .net core available
3. edit deploy.ps1 and populate with a SQL-auth user/password; otherwise if using windows auth you can probably remove lines 21/22 and use that
4. run .\deploy.ps1, and then run it again (so we can use the pre-deploy script).  It will fail both times, but you can see the results with the following query:

```
SELECT *
FROM [MsBuildSqlTest].[dbo].[PreDeployTable]

SELECT *
FROM [MsBuildSqlTest].[dbo].[PostDeployTable]
```

5. if desired, you can comment out some of the other SQLCMD variables (TestValue6, TestValue7, TestValue8, TestValue99) and the corresponding lines in the pre/post-deploy scripts.  Also you can uncomment TestValue1/TestValue2 to see the compilation errors

ErikEJ:
fixes #701


jmezach (approved):
LGTM

jeffrosenberg (approved):

PR #711, opened 2025-02-08
Create GitHub action to keep ManagedBatchParser dependency up to date

### LINKED ISSUE #681: Ensure ManagedBatchParser releases is monitored automatically
@jeffrosenberg FYI

jeffrosenberg:
This adds a new scheduled workflow to check for new versions of ManagedBatchParser. It checks the GitHub release metadata to get the latest version, and compares it to the current dependency version. If the two don't match, it downloads the new one and creates a PR to update the version.

There are a few decisions I've made in this PR that I don't feel strongly about, but I had to pick _something_ to get the PR out. I'll highlight those for further discussion; they should be easy to change if we want different approaches.

Resolves #681


.github/workflows/update-managed-batch-parser.yml
```diff
@@ -0,0 +1,142 @@
...
+      # (false if the versions match, true if they don't).
+      # The create_pr job is conditional on this value
+    - id: compare_versions
+      name: Compare versions and determine if update is needed
+      env:
+        CURRENT: ${{needs.get_versions.outputs.current}}
+        LATEST: ${{needs.get_versions.outputs.latest}}
+      run: |
+        needs_update=false
+        [[ "$CURRENT" != "$LATEST" ]] && needs_update=true
```
jeffrosenberg, line 86: The only comparison I'm doing is checking whether the versions are exactly equal. The [sqltoolsservice repo](https://github.com/microsoft/sqltoolsservice) isn't using semantic versioning, or I might have tried to break the version down into major/minor/patch and let us configure what types of changes we wanted to trigger an update. If anyone thinks we should make this version check more robust, I'm open to that.

src/DacpacTool/Microsoft.SqlTools.ManagedBatchParser/version.txt
jeffrosenberg, line 1: I chose the simplest way to track the version of ManagedBatchParser that I could think of: I moved the dll into its own directory, and added a text file with the version number. I wondered if I should extend NBGV to track this version, but it didn't look to me like that was exactly its intended use case. Open to other ideas about how to track this.

.github/workflows/update-managed-batch-parser.yml
```diff
@@ -0,0 +1,142 @@
+name: Update ManagedBatchParser
+
+on:
+  schedule:
+  - cron: "50 11 * * 0" # 11:50 am UTC, every Sunday
```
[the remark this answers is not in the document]
ErikEJ: Weekly seems plenty to me!

.github/workflows/update-managed-batch-parser.yml
ErikEJ: With the current low number of releases, any update seems fine, and keeps things simple

ErikEJ (approved):
LGTM

jeffrosenberg:
> Looking at the release assets here https://github.com/microsoft/sqltoolsservice/releases/tag/5.0.20250206.1, I am not sure that we can use any of the binaries from the .tar files, as they all seem to be very platform specific.
> 
> But getting notified via a PR seems a good approach, to indicate that we should consider if an update is needed!

Yeah, this would be worth testing by someone running a Windows machine. On my own Linux machine, I was able to check out a PR branch during testing and get a successful build, but maybe that's just because I'm on Linux. Someone could try checking out [this branch on my fork](https://github.com/jeffrosenberg/MSBuild.Sdk.SqlProj/tree/actions/managed-batch-parser-5.0.20250206.1) to see if it will build for you.

ErikEJ:
@jeffrosenberg I pulled down you branch, built DacpacTool and got this warning:

`1>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(2413,5): warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "Microsoft.SqlTools.ManagedBatchParser", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.
`

.github/workflows/update-managed-batch-parser.yml
```diff
@@ -0,0 +1,142 @@
...
+      name: Get currently-used version from version.txt
+      working-directory: repo
+      run: echo "current=$(cat src/DacpacTool/Microsoft.SqlTools.ManagedBatchParser/version.txt)" >> "$GITHUB_OUTPUT"
+
+      # Get sqltoolsservice release metadata,
+      # store in a file for later processing using JQ
+    - id: curl_sqltoolsservice
+      name: Get latest release metadata for sqltoolsservice
+      working-directory: working
+      run: curl https://api.github.com/repos/microsoft/sqltoolsservice/releases/latest | tee curl.txt
```
jmezach, line 43: Do we need to worry about any rate limiting issues here? I guess not since we're running this once a week, I guess I'm just curious ;).

jeffrosenberg: Per [these docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28), "The primary rate limit for unauthenticated requests is 60 requests per hour". :slightly_smiling_face:

jeffrosenberg:
> @jeffrosenberg I pulled down you branch, built DacpacTool and got this warning:
> 
> `1>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(2413,5): warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "Microsoft.SqlTools.ManagedBatchParser", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project. `

@ErikEJ okay, that confirms your concern. Is there a better source I should be using for this file? Where did you use to source the ManagedBatchParser DLL file from? I'd note that [our build runs on Ubuntu](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/master/.github/workflows/main.yml#L15-L19), but is there a version of the binary that isn't platform-specific?

ErikEJ:
I had to build it my self 😢

jeffrosenberg:
> I had to build it my self 😢

Doesn't that still end up specific to the platform you built it on? I would argue that since our build is on Ubuntu, we should be fine to use the Ubuntu-flavored release. But I'll acknowledge that I don't perfectly understand how this works. 

If it would be preferable, I could definitely adapt this automation to clone and build the sqltoolsservice repo, so we don't have to have a human do it each time a new release is available. But I'd still need to choose an agent for that build, so based on my limited understanding, that would still end up being "Ubuntu flavored" by virtue of building on an Ubuntu agent.

ErikEJ:
I am sure it will be perfectly fine with the Ubuntu build, there is no unmanaged code involved in the part of the SqlToolsService

.github/workflows/update-managed-batch-parser.yml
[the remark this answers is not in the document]
jmezach: Agreed

.github/workflows/update-managed-batch-parser.yml
jmezach: Let's just keep it simple :).

PR #727, opened 2025-04-06
Code health - Enable nullable warnings

ErikEJ:


src/DacpacTool/Extensions.cs
```diff
@@ -141,8 +141,27 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
 
             var crossPlatformPath = referencePath.Replace('\\', '/');
 
-            var customData = Activator.CreateInstance(Type.GetType("Microsoft.Data.Tools.Schema.SchemaModel.CustomSchemaData, Microsoft.Data.Tools.Schema.Sql"), "Reference", "SqlSchema");
+            var customType = Type.GetType("Microsoft.Data.Tools.Schema.SchemaModel.CustomSchemaData, Microsoft.Data.Tools.Schema.Sql");
```
jmezach, line 144: Looking at this again now I'm wondering if we should do this statically, instead of doing this every time we add a reference. We could add a static constructor that resolves these types and methods.

ErikEJ: Not 100% sure, but attempted an update for this

src/DacpacTool/ScriptParser.cs
```diff
@@ -69,7 +69,15 @@ BatchParserAction ICommandHandler.Include(TextBlock includeBlock, out TextReader
             if (!Path.IsPathRooted(includedFileName))
             {
                 var position = lineInfo.GetStreamPositionForOffset(0);
-                includedFileName = Path.Combine(Path.GetDirectoryName(position.Filename), includedFileName);
+
+                var projectDir = Path.GetDirectoryName(position.Filename);
```
jeffrosenberg, line 73: If we're saying `Unable to resolve path for included file`, I think we should test the file itself, instead of the directory. Couldn't we have a scenario here where we validate that `projectDir` exists, but `includedFileName` _doesn't_ exist? Or am I misunderstanding what this is checking?

src/DacpacTool/Extensions.cs
```diff
@@ -16,6 +16,39 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
     public static class Extensions
 #pragma warning restore CA1724 // Type names should not match namespaces
     {
+        private static Type GetCustomSchemaDataType()
```
jmezach, line 19: I was thinking more about a static constructor that resolves these types and methods and puts them into static fields that can then be used

jmezach: Hmm, interesting, not allowed by what?

ErikEJ:
@jeffrosenberg @jmezach Comments adressed

src/DacpacTool/Extensions.cs
```diff
@@ -212,10 +251,22 @@ private static string EnsureIsDelimited(string name)
         {
             if (_ensureIsDelimitedMethod == null)
             {
-                _ensureIsDelimitedMethod = Type.GetType("Microsoft.Data.Tools.Schema.Common.FileUtils, Microsoft.Data.Tools.Utilities")
+                var method = Type.GetType("Microsoft.Data.Tools.Schema.Common.FileUtils, Microsoft.Data.Tools.Utilities");
```
jmezach, line 254: Perhaps rename this to variable to `fileUtilsType` as it is now no longer a method?

ErikEJ:
@jeffrosenberg @jmezach Do you have time for another look?

jeffrosenberg (approved):

jmezach (approved):
LGTM

PR #738, opened 2025-04-25
Fix logging of DacModelExceptions

### LINKED ISSUE #734: Issues with the Error List pane in Visual Studio
I'm noticing a couple of issues with the Error List pane in Visual Studio.

Image

For one, when you double click on an error it doesn't take you to the file.

Secondly, the Path column doesn't show the correct value. It's showing the path to the root of the database project, not the path to the erroring SQL script file. For example, in the screenshot above you can see the DistressSource.sql script shows the path as: C:\Dev\RPR\Databases\StaticData, however the build output shows the correct path:

> 1>Adding C:\Dev\RPR\Databases\StaticData\reference\StaticData\DistressSource.sql to the model
> 1>DistressSource.sql(1,1): error SQL46010: Incorrect syntax near '\'.

ErikEJ:
fixes #734

### DIFF
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -82,10 +82,11 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         {
             ArgumentNullException.ThrowIfNull(exception);
 
-            var stringBuilder = new StringBuilder();
+            var lines = new List<string>();
 
             foreach (var modelError in exception.Messages)
             {
+                var stringBuilder = new StringBuilder();
                 stringBuilder.Append(fileName);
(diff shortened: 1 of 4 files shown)


src/DacpacTool/Extensions.cs
```diff
@@ -82,10 +82,11 @@ public static string Format(this DacModelException exception, string fileName)
         {
             ArgumentNullException.ThrowIfNull(exception);
 
-            var stringBuilder = new StringBuilder();
+            var lines = new List<string>();
```
jmezach, line 85: Can't we just call `AppendLine()` on the StringBuilder?

jeffrosenberg (approved):

PR #746, opened 2025-05-16
enable net8.0 target framework

### LINKED ISSUE #745: Support .NET 8 only analyzers as references
DacFX 170 does no longer support netstandard2.0 - wonder if we can make this work with new analyzer packages during build?

ErikEJ:
fixes #745


jmezach (approved):
LGTM

PR #751, opened 2025-05-22
Make StaticCodeAnalysis.SuppressMessages.xml support cross platform

### LINKED ISSUE #749: Support OS-agnostic file paths (or wildcards) in StaticCodeAnalysis.SuppressMessages.xml
I’ve discovered that when I build my database project on Windows vs. Linux, per-file rule suppressions in StaticCodeAnalysis.SuppressMessages.xml only work if I use the OS-specific path separator (\ on Windows, / on Linux). That means I have to duplicate every <SuppressedFile> entry to get the same rule suppressed on both platforms.

It would be much more convenient if the SDK either:
- Normalized path separators before matching, so a single entry works everywhere, or
- Accepted simple globs or regex (e.g. wes[/\\]Stored Procedures[/\\]StoGetStatusMetricCounts\.sql)

___

### Current Workaround:
```xml
<?xml version="1.0" encoding="utf-8"?>
<StaticCodeAnalysis version="2" xmlns="urn:Microsoft.Data.Tools.Schema.StaticCodeAnalysis">

    <!-- MSBuild.Sdk.SqlProj’s per‐file suppression is literally string-matching the FilePath you give it,
         so a backslash only matches on Windows and a forward slash only matches on Linux.
         The simplest cross-platform workaround is to just include both variants in your StaticCodeAnalysis.SuppressMessages.xml:
    -->
    
    <!-- Windows build (backslashes) for local building -->
    <SuppressedFile FilePath="wes\Stored Procedures\StoGetStatusMetricCounts.sql">
        <SuppressedRule Category="Smells" RuleId="SML038" />
    </SuppressedFile>

    <!-- Linux build (forward slashes) for CI building -->
    <SuppressedFile FilePath="wes/Stored Procedures/StoGetStatusMetricCounts.sql">
        <SuppressedRule Category="Smells" RuleId="SML038" />
    </SuppressedFile>
</StaticCodeAnalysis>
```
___

### Desired Behavior:
- A single <SuppressedFile FilePath="wes/Stored Procedures/StoGetStatusMetricCounts.sql"> entry should suppress on both Windows and Linux, or
- <SuppressedFile> entries should accept wildcard/regex patterns so one rule can cover both path formats.
___

Thanks for considering this improvement—cross-platform builds would be much simpler without duplication!

ErikEJ:
fixes #749

### DIFF
--- a/src/DacpacTool/PackageAnalyzer.cs
+++ b/src/DacpacTool/PackageAnalyzer.cs
@@ -74,7 +74,7 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 {
                     service.SetProblemSuppressor(p => 
                         suppressedProblems.Any(s => s.Rule.RuleId == p.Rule.RuleId
-                            && Path.Combine(projectDir, s.SourceName) == p.Problem.SourceName)
+                            && Path.Combine(projectDir, s.SourceName).Replace('\\', '/') == p.Problem.SourceName.Replace('\\', '/'))
                         || _ignoredRules.Contains(p.Rule.RuleId)
                         || _ignoredRuleSets.Any(s => p.Rule.RuleId.StartsWith(s, StringComparison.OrdinalIgnoreCase)));
                 }
--- a/test/DacpacTool.Tests/Suppression/StaticCodeAnalysis.SuppressMessages.xml
+++ b/test/DacpacTool.Tests/Suppression/StaticCodeAnalysis.SuppressMessages.xml
@@ -1,6 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <StaticCodeAnalysis version="2" xmlns="urn:Microsoft.Data.Tools.Schema.StaticCodeAnalysis">
   <SuppressedFile FilePath="proc1.sql">
-	<SuppressedRule Category="Microsoft.Rules.Data" RuleId="SR0001" />
+	  <SuppressedRule Category="Microsoft.Rules.Data" RuleId="SR0001" />
+  </SuppressedFile>
+  <SuppressedFile FilePath="Folder\proc3.sql">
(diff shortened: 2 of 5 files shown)


src/DacpacTool/PackageAnalyzer.cs
```diff
@@ -74,7 +74,7 @@ public void Analyze(TSqlModel model, FileInfo outputFile, FileInfo[] analyzers)
                 {
                     service.SetProblemSuppressor(p => 
                         suppressedProblems.Any(s => s.Rule.RuleId == p.Rule.RuleId
-                            && Path.Combine(projectDir, s.SourceName) == p.Problem.SourceName)
+                            && Path.Combine(projectDir, s.SourceName).Replace('\\', '/') == p.Problem.SourceName.Replace('\\', '/'))
```
jmezach, line 77: Aren't there existing constants for these characters that we could use?

test/DacpacTool.Tests/Suppression/StaticCodeAnalysis.SuppressMessages.xml
```diff
@@ -1,6 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <StaticCodeAnalysis version="2" xmlns="urn:Microsoft.Data.Tools.Schema.StaticCodeAnalysis">
   <SuppressedFile FilePath="proc1.sql">
-	<SuppressedRule Category="Microsoft.Rules.Data" RuleId="SR0001" />
+	  <SuppressedRule Category="Microsoft.Rules.Data" RuleId="SR0001" />
```
jmezach, line 4: Indentation is a bit weird here

jmezach (approved):
LGTM

jeffrosenberg (approved):

PR #775, opened 2025-08-03
WIP: Better incremental build

ErikEJ:
DRAFT!

This is my uninformed attempt.

Can see that the hash value changes, if the project files change (removal/addition)

But a change in the hash does not currently trigger a build

The MS SDK also has:

```xml
<PropertyGroup>
    <SqlBuildDependsOn>$(SqlBuildDependsOn);_GenerateSqlBuildDependencyCache</SqlBuildDependsOn>
  </PropertyGroup>
```


ErikEJ:
@jmezach 

The issue is with the default `dotnet build` process (or build in VS). 

If you remove a file, you would expect a build to trigger and produce a new .dacpac, but that does not currently happen, so you have to force a build, with for example `dotnet build --no-incremental`

The idea is to have the cache file change when a file is removed, so a build is triggered in that case.

PR #789, opened 2025-09-23
Use Trusted Publishing

### LINKED ISSUE #787: Use Trusted Publishing
https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing

jmezach:
This enables the use of Trusted Publishing for publishing our NuGet packages to NuGet.org enhancing security.

Fixes #787

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,9 @@ on:
     branches: [ master ]
   workflow_dispatch:      
 
+permissions:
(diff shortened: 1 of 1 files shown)


ErikEJ:
@jmezach Maybe permission need to be added in two places, looking at the sample here?

https://devblogs.microsoft.com/dotnet/enhanced-security-is-here-with-the-new-trust-publishing-on-nuget-org/#minimal-github-actions-example-⚙️

Issue #733, opened 2025-04-24
Container build support


jmezach:
As discussed in https://github.com/CommunityToolkit/Aspire/issues/595 it would be beneficial if we could come up with a way to support `dotnet publish /t:PublishContainer` for SQL Database Projects. The output would be a container that would contain both `sqlpackage` and the output `.dacpac`. Running the container would essentially be running `sqlpackage` with some pre-populated arguments (such as the path to the `.dacpac`) and of course the ability to pass additional arguments as needed (such as the target connection string). Such a container could then be used in a CI/CD pipeline to deploy the database schema to a SQL Server.

ErikEJ:
@dzsquared any plans to publish a sqlpackage container? We need it 😅

ErikEJ:
Some example (using the full MSSQL container, I guess we do not want to do that!) https://github.com/jcgillespie/mssql-sqlpackage-docker

Some installation advice https://learn.microsoft.com/en-us/sql/tools/sqlpackage/sqlpackage-download?view=sql-server-ver17#installation-cross-platform - both for using with an image with .NET SDK and without

jmezach:
I would assume that we could use the .NET runtime image as a base and then add sqlpackage on top right?

ErikEJ:
Sounds amazing - but we are now back to the lack of an existing .NET container with sqlpackage.

Could we use some of the effort in the abandoned community PR for the download?

I assume we would prefer that over publishing and supporting our "own" inage?

jmezach:
@ErikEJ We could, but then you would be publishing a container that has a lot of stuff that you don't really need. Some folks might not like that and it can also be problematic from a security point of view. The fewer things are in that container the better imho.

Fun fact is that I did download the latest SqlPackage for Linux and put that into the container and now I'm getting this when I run it:

```
rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
```

Haven't looked into it any further yet. Might be some weird thing with arm64 vs x64 or something.

ErikEJ:
@jmezach I think the plain runtime container will not be enough to run sqlpackage - I have had similar issues before

see https://learn.microsoft.com/en-us/sql/tools/sqlpackage/sqlpackage-download?view=sql-server-ver17#linux-x64

and

https://learn.microsoft.com/en-us/sql/tools/sqlpackage/sqlpackage-download?view=sql-server-ver17#linux 

So we are back at our own custom container ? 😢

[...]

PR #799, opened 2025-10-20
feat(build): container publish for image with dacpac and sqlpackage

nmummau:
## Proposal to resolve issue #733 

This PR provides a new build target "PublishContainer" which will create a docker image that contains both your project's dacpac and sqlpackage. This image can then be run to publish your dacpac to a target database.

Demo video for the PR feature: https://youtu.be/NaCNs0OOUbM

## How I've tested this
I've been building and packing the SDK with 
```sh
dotnet build -c Release &&
dotnet pack MSBuild.Sdk.SqlProj.csproj -c Release
```

I then copy the nupkg...
`C:\code\MSBuild.Sdk.SqlProj\src\MSBuild.Sdk.SqlProj\bin\Release\MSBuild.Sdk.SqlProj.3.3.0-beta.16.g87c1e17ddb.nupkg`

...to a directory I use as a local NuGet feed
`C:\LocalFeed`

Then in the project I want to use this, for me this is `c/code/wes/inventory/src`
```sh
dotnet nuget locals all --clear
```

🧱 Then build with the PublishContainer target
```sh
dotnet build ./WES.Inventory.Database/WES.Inventory.Database.csproj \
-t:PublishContainer \
-c Release \
-v:m
```

Due to the name of the project I'm using this with my resulting docker image is named `wes.inventory.database-publisher:release`

You can look into the files within the image
```sh
MSYS_NO_PATHCONV=1 MSYS2_ARG_CONV_EXCL="*" \
docker run --rm --entrypoint sh wes.inventory.database-publisher:release \
-lc "ls -l /work"
```
results:
```
-rwxr-xr-x 1 root root  14442 Oct 19 15:31 WES.Inventory.Database.dacpac
-rwxr-xr-x 1 root root 180940 Apr 11  2025 master.dacpac
```

🎊 Finally you can run the image (git bash) to publish your dacpac to the target database
```sh
MSYS_NO_PATHCONV=1 \
docker run --rm wes.inventory.database-publisher:release \
-TargetServerName:host.docker.internal,14333 \
-TargetDatabaseName:WES_001_inv \
-TargetUser:sa \
-TargetPassword:YourPassword123! \
-TargetTrustServerCertificate:true \
"/p:DropObjectsNotInSource=True" "/p:BlockOnPossibleDataLoss=False"
```

results
```sh
...
Publishing to database 'WES_001_inv' on server 'host.docker.internal,14333'.
Initializing deployment (Start)
Initializing deployment (Complete)
Analyzing deployment plan (Start)
Analyzing deployment plan (Complete)
Updating database (Start)
Creating database WES_001_inv...
Creating Schema [wes]...
...
```

### DIFF
--- a/src/DacpacTool/DacpacTool.csproj
+++ b/src/DacpacTool/DacpacTool.csproj
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SqlServer.DacFx" Version="$(DacFxVersion)" />
+    <PackageReference Include="Microsoft.SqlServer.DacFx" Version="170.2.70" />
     <PackageReference Include="NuGet.Versioning" Version="6.14.0" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageReference Include="System.CommandLine.NamingConventionBinder" Version="2.0.0-beta4.22272.1" />
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
@@ -10,7 +10,7 @@
     <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)' != 'Current'">$(MSBuildAllProjects);$(MsBuildThisFileFullPath)</MSBuildAllProjects>
     <SqlServerVersion>Sql150</SqlServerVersion>
     <TargetExt>.dacpac</TargetExt>
-    <DacFxVersion>170.2.70</DacFxVersion>
+    <SqlPackageVersion>170.2.70</SqlPackageVersion>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -323,9 +323,9 @@
 
     <!--
       Passed to the Dockerfile as build-arg SQLPACKAGE_VERSION.
-      Using DacFxVersion which is set in Sdk.props so that we keep DacFx and SqlPackage in sync
+      Using SqlPackageVersion which is set in Sdk.props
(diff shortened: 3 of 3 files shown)


nmummau:
> @nmummau Why do we need to have the sqlpackage version involved?
> 
> Could we not just use:
> 
> `dotnet tool install -g Mircosoft.Sqlpackage`

Short answer: we can use the .NET global tool, but the current approach (pulling the NuGet nupkg/platform zip) is more deterministic and works on the slimmer base image without extra fuss.

`dotnet tool install -g ...` requires the .NET SDK as the dotnet/runtime image doesn't include the tooling needed to install global tools

Also, pinning to a `SQLPACKAGE_VERSION` will give us deterministic versioning. Even if we used the global tool we should probably pin the version (`dotnet tool install -g Microsoft.SqlPackage --version 162.3.15`). Because If you leave it floating, upgrades happen silently, which can introduce subtle failures.

If you think we should use the .NET SDK with the global tool, instead of what I have, I can give that a shot and see what I come up with. Thought?

nmummau:
@ErikEJ this is aside, but I wanted to show you a PR (with another guide video) I created for a different project where I implemented some Docker tooling for the purpose of deploying to a containerized database using a `.dacpac` with sqlpackage. This mechanism works, but should show off to you why it would be very nice if the SDK could provide that functionality out of the box

https://github.com/Eventuous/eventuous/pull/452

jmezach:
@nmummau I really see the value of having something like this in the box. Having the ability to just publish a database project as a container that when run would deploy it to a target database can help in a lot of scenario's.

While I appreciate the effort you've put into this and I'm not entirely sure this is the right approach to implementing this feature. In its current form it seems that we're running `docker` on the command line. This obviously requires some form of Docker runtime to be installed on the machine. Looking at the built-in support for the .NET SDK it seems that this is not required there, see https://learn.microsoft.com/en-us/dotnet/core/containers/sdk-publish#prerequisites. I think that if it works like that this feature would be even more powerful.

ErikEJ:
@nmummau Are you planning to do more work on this?

Issue #798, opened 2025-10-19
Contributing guide seems to be missing something


nmummau:
Am I doing something wrong with the test projects? The CONTRIBUTING.md says this:
>You can then build any of the test projects in the test folder to try out your changes without having to build the SDK package and push it a NuGet package feed.

However for me that doesn't seem true because, for example, if I open `test\TestProjectWithPackageReference\TestProjectWithPackageReference.csproj` and build I get the following error
```
Unable to find package TestProject with version (>= 1.2.0-beta.9.gf2d69b1c16)
  - Found 1 version(s) in nuget.org [ Nearest version: 1.0.0 ]
  - Found 1 version(s) in Clark Feed [ Nearest version: 1.0.0 ]
  - Found 0 version(s) in C:\code\MSBuild.Sdk.SqlProj\test\TestProject\bin\Debug
  - Found 0 version(s) in Local Feed
  - Found 0 version(s) in Microsoft Visual Studio Offline Packages
 ```

If I try to open the `test\TestProjectWithSDKRef\TestProjectWithSDKRef.csproj` and build I get the following error:
```
The SDK 'MSBuild.Sdk.SqlProj/#{NBGV_NuGetPackageVersion}#' specified could not be found.
```

PR #806, opened 2025-11-10
Add some comments to the test projects

jmezach:
As discussed in #798 this adds some comments to the test projects so explain how they can be build locally if needed.


ErikEJ (approved):

jeffrosenberg (approved):
`CONTRIBUTING.md` says:

> You can then build any of the test projects in the test folder to try out your changes without having to build the SDK package and push it a NuGet package feed.

I might also recommend adding a note that says "further information on testing can be found within the test projects themselves."

While we're at it, we could update the references to the .NET versions in `CONTRIBUTING.md` as well -- or maybe just remove direct version references so we don't have to keep updating it :slightly_smiling_face:

CONTRIBUTING.md
```diff
@@ -25,12 +25,12 @@ If a maintainer asks you to "rebase" your PR, they're saying that a lot of code
 # Getting started
 In order to get started contributing code, make sure you have the following installed on your machine:
 
-* .NET 6 SDK
+* .NET SDK 6.0 or later (you can get it from [here](https://dotnet.microsoft.com/en-us/download))
```
ErikEJ: ".NET SDK 6.0 or later" => ".NET SDK"

PR #807, opened 2025-11-10
Hide bin folder in VS

### LINKED ISSUE #805: Solution explorer shows bin folder for this project type
Could be related to support for netx.0 taget framework (does not occur with .net standard 2.1 target)

Could we add this to the .props file?

` <None Remove="bin/**" />`

ErikEJ:
fixes #805

### DIFF
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
@@ -44,6 +44,7 @@
     </Content>
     <Content Remove="Pre-Deployment/**/*.sql" />
     <Content Remove="Post-Deployment/**/*.sql" />
+    <!-- GenerateScript adds a .sql file here, and it should not be Content in the .dacpac -->
     <Content Remove="bin/**/*.sql" />
     <None Remove="bin/**" />
     <Content Remove="obj/**/*.sql" />


src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
```diff
@@ -45,6 +45,7 @@
     <Content Remove="Pre-Deployment/**/*.sql" />
     <Content Remove="Post-Deployment/**/*.sql" />
     <Content Remove="bin/**/*.sql" />
```
jeffrosenberg, line 48: Do we still need this line if we have the broader rule for `bin/**`?

jeffrosenberg (approved):

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.props
ErikEJ: No sure, this line relates to compilation items? @jmezach  ?

jmezach: I think this was added to ensure that if you use GenerateScript that the resulting script file is not also included as a script to include in the `.dacpac`. I think it makes sense to leave it, but perhaps add a comment explaining why?

jmezach (approved):
LGTM

PR #832, opened 2026-01-06
Remove hardcoded secret

jmezach:
This removes a hardcoded secret from the GitHub Actions pipeline. This was brought to my attention by a security researcher recently. Although the password is being used only for an ephemeral container I think it is good idea to follow best practices around security.


jmezach:
@ErikEJ @jeffrosenberg I would like to get this merged sooner rather than later. I can bypass rules if need be.

PR #834, opened 2026-01-06
Improve Publish experience in VS

### LINKED ISSUE #833: Improve VS Publish experience
If this gets added as a default when running under VS, the Publish experience will be so much better - right menu Publish "Just works"

`<TargetServerName>(localdb)\mssqllocaldb</TargetServerName>`

@jmezach is it possible to detect that the props are running under VS?

ErikEJ:
fixes #833

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -318,7 +318,7 @@ jobs:
 
     # Upload binary log
     - name: upload
-      if: always()
+      if: failure()
       uses: actions/upload-artifact@v6
       with:
         name: binary-log-publish


ErikEJ:
@jmezach Good idea, how do I do that?

jmezach:
@ErikEJ Looks like we're already doing it (see https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/166c5da7f293672bad9deed93ca951cbd2d91ec9/.github/workflows/main.yml#L319-L324), but only if the previous step was successful. I guess we can change the condition to always do this, even if the previous step failed

ErikEJ:
Looks very promising!

ErikEJ:
Verfied by Publishing TestProject.csproj from VS to LocalDB

.github/workflows/main.yml
```diff
@@ -318,6 +318,7 @@ jobs:
 
     # Upload binary log
     - name: upload
+      if: always()
```
jeffrosenberg: Just confirming whether we want to keep this? Seems reasonable to me to keep it, but I'm not sure whether the intention was just for debugging the failing build in this PR

ErikEJ: Good point! I think it makes sense to keep it - @jmezach ?

jmezach: The binary log includes environment variables, including our recently added secret, so maybe we should only upload it in case of a failure. If it succeeds I don't really care that much about the binary log.

PR #835, opened 2026-01-11
Update to use System.Commandline 2.0.1

### LINKED ISSUE #833: Improve VS Publish experience

ErikEJ:
I got help from Copilot with this one, so let's see how it goes!

fixes #833


ErikEJ:
The aliases are missing, working on it.

PR #836, opened 2026-01-12
Update System.CommandLine 

### LINKED ISSUE #830: System.Commandline update/migration
System.Commandline is now finally released, but we lost NaminngConventionBinder on the way. which means we need to add a lot of mapping code ourselves as I understand it.

(I tried to update once, but it was too complex to just do)

Alternatives include:

Stay on beta 4

Update and add the extra mapping code

Use another command line parser

@jeffrosenberg @jmezach Thoughts?

ErikEJ:
fixes #830

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -217,7 +217,7 @@ jobs:
         echo "Waiting for SQL Server to start..."
         sql_ready=false
         for i in {1..30}; do
-          if docker exec sqlserver /opt/mssql-tools18/bin/sqlcmd -C -S localhost -U sa -P ${{ secrets.SQL_SA_PASSWORD }} -Q "SELECT 1" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


ErikEJ:
@jmezach Looks like the failure here could be related to your recent security / pw related update?

jmezach:
@ErikEJ Weird, it looks almost like there's no value for the secret at all. I've updated the secret with a new value, maybe that resolves the issue?

ErikEJ:
@jmezach Yeah. looks like no value is passed through - is there soem scoping / access to the secret involved?

ErikEJ:
@jmezach Maybe a variable is a better approach?

ErikEJ:
@jmezach I probably could, but never do.

ErikEJ:
@jmezach @jeffrosenberg Replace by #837

PR #840, opened 2026-01-15
Add container publish support

### LINKED ISSUE #733: Container build support

jmezach:
This adds support for publishing an MSBuild.Sdk.SqlProj project as a runnable container image that when run deploys the database schema using `sqlpackage.exe`.

There's a couple of things to note about the current implementation:
- I've renamed the existing Publish target as PublishDatabase. This was necessary to avoid running both the "deploy-to-sqlserver" flow and "publish-container-image" flow with the defacto standard command of `dotnet publish /t:PublishContainer`. **This is a BREAKING CHANGE**.
- I've defaulted to building container images as `linux-x64` since that's the only architecture on which `sqlpackage` is supported
- We don't support publishing Windows containers with this flow.
- We download sqlpackage from the pseudo-link `https://aka.ms/sqlpackage-linux` which always points to the latest release. Do we want this?
- We download sqlpackage to a temporary folder and then extract it into the $(PublishDir) so that it ends up in the container image. We'll probably want to avoid downloading it over and over and I'm not yet sure if the current logic covers that

Fixes #733

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,7 +539,11 @@ Most of those properties are simple values (like booleans, strings and integers)
 | SqlCommandVariableValues | | These should not be set as a Property, but instead as an ItemGroup as described [in this section](#sqlcmd-variables) |
 
 ### Publishing as a container image
-From version 4.0.0 of MSBuild.Sdk.SqlProj we now support publishing your database project as a runnable container image. The image will contain both SqlPackage and the .dacpac file. This allows you to run the image anywhere a container can be executed, making it ideal for CI/CD pipelines and other automated deployment scenarios. To publish your project as a container image, use the following command:
+From version 4.0.0 of MSBuild.Sdk.SqlProj we now support publishing your database project as a runnable container image. The image will contain both SqlPackage and the .dacpac file. This allows you to run the image anywhere a container can be executed, making it ideal for CI/CD pipelines and other automated deployment scenarios.
+
+> Note: By default the published container will contain the latest version of SqlPackage available at the time of publishing. If you want to pin the specific version of SqlPackage used in the container you can set the `SqlPackageDownloadUrl` property in your project file to point to the specific version you want, ie: `https://go.microsoft.com/fwlink/?linkid=2338525` for version 170.2.70. You can find the appropriate download links on the [SqlPackage release notes](https://learn.microsoft.com/en-us/sql/tools/sqlpackage/release-notes-sqlpackage) page. Please make sure to use the link for the Linux .NET 8 version of SqlPackage.
+
+To publish your project as a container image, use the following command:
 
 ```bash
 dotnet publish /t:PublishContainer
--- a/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
+++ b/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
@@ -330,11 +330,11 @@
   <Target Name="PrepareContainer" DependsOnTargets="CopyDacpacs">
     <PropertyGroup>
       <TempStagingFolder>$([System.IO.Path]::GetTempPath())</TempStagingFolder>
-      <SqlPackageStagingFolder>$([System.IO.Path]::Combine('$(TempStagingFolder)', 'sqlpackage-linux'))</SqlPackageStagingFolder>
+      <SqlPackageStagingFilename>$([System.IO.Path]::Combine('$(TempStagingFolder)', 'sqlpackage-linux.zip'))</SqlPackageStagingFilename>
       <SqlPackageExtractPath>$([System.IO.Path]::Combine('$(PublishDir)', 'sqlpackage'))</SqlPackageExtractPath>
     </PropertyGroup>
-    <DownloadFile SourceUrl="$(SqlPackageDownloadUrl)" DestinationFolder="$(TempStagingFolder)" Condition="!Exists('$(SqlPackageExtractPath)')" />
-    <Unzip SourceFiles="$(SqlPackageStagingFolder)" DestinationFolder="$(PublishDir)" Condition="!Exists('$(SqlPackageExtractPath)')" />
+    <DownloadFile SourceUrl="$(SqlPackageDownloadUrl)" DestinationFolder="$(TempStagingFolder)" DestinationFilename="sqlpackage-linux.zip" Condition="!Exists('$(SqlPackageExtractPath)')" />
+    <Unzip SourceFiles="$(SqlPackageStagingFilename)" DestinationFolder="$(PublishDir)" Condition="!Exists('$(SqlPackageExtractPath)')" />
     <Copy SourceFiles="@(CopyDacpacFiles)" DestinationFolder="$(PublishDir)" />
   </Target>
 </Project>


jmezach:
Took me more attempts than I was hoping for, but I managed to get it all working ;). I've also written up some docs. Would appreciate a review here @ErikEJ and @jeffrosenberg. I have a feeling this feature could be a game changer :).

README.md
```diff
@@ -531,12 +538,50 @@ Most of those properties are simple values (like booleans, strings and integers)
...
+  </PropertyGroup>
+</Project>
+```
+
+> Note: In older versions of MSBuild.Sdk.SqlProj we've recommended to use netstandard2.0 as the target framework. This has since been changed to net8.0 (or higher). In order to publish container images you'll need to be targeting net8.0 or higher and use at least version 8.0 of the .NET SDK.
+
+Once the container image is published, you can run it using the following command:
+
+```bash
+docker run my-database-image:v1.0.0 --rm /TargetServerName=<server> /TargetDatabaseName=<database> /TargetUser=<username> /TargetPassword=<password>
```
ErikEJ: Maybe just use TargetConnectionString?

jmezach: Sure, that'll also work and less for us to maintain ;)

README.md
```diff
@@ -531,12 +538,50 @@ Most of those properties are simple values (like booleans, strings and integers)
...
 | SqlCommandVariableValues | | These should not be set as a Property, but instead as an ItemGroup as described [in this section](#sqlcmd-variables) |
 
+### Publishing as a container image
+From version 4.0.0 of MSBuild.Sdk.SqlProj we now support publishing your database project as a runnable container image. The image will contain both SqlPackage and the .dacpac file. This allows you to run the image anywhere a container can be executed, making it ideal for CI/CD pipelines and other automated deployment scenarios. To publish your project as a container image, use the following command:
+
+```bash
+dotnet publish /t:PublishContainer
+```
+
+By default, the image will be tagged with the name of your project. You can customize the image name and tag by setting the `ContainerImageName` and `ContainerImageTag` properties on the command line. For example:
```
baronfel: Please use ContainerRepository instead of ContainerImageName - repository is the standard term in use here across the containers tech landscape. We only have the image name property for back compat at this point, and it may be deleted in the future.

README.md
```diff
@@ -531,12 +538,50 @@ Most of those properties are simple values (like booleans, strings and integers)
...
+  <PropertyGroup>
+    ...
+    <ContainerImageName>my-database-image</ContainerImageName>
+    <ContainerImageTag>v1.0.0</ContainerImageTag>
+    ...
+  </PropertyGroup>
+</Project>
+```
+
+> Note: In older versions of MSBuild.Sdk.SqlProj we've recommended to use netstandard2.0 as the target framework. This has since been changed to net8.0 (or higher). In order to publish container images you'll need to be targeting net8.0 or higher and use at least version 8.0 of the .NET SDK.
```
baronfel: I believe you should be able to continue to use NS2. 0 if you like - the SDK Container targets use the TFM to help determine the base image to use, but since you're controlling other aspects of the image build you have to power to control that logic as well.

README.md
jmezach: This was actually suggested by Copilot and I didn't really think about it, but good catch. I'll be sure to update that.

README.md
jmezach: I tried this on one of our test projects that still had netstandard2.0 as the TFM and I get the following error:

```
error CONTAINER1015: Unable to access the repository 'dotnet/runtime' at tag '2.0' in the registry ''. Please confirm that this name and tag are present in the registry.
```

I guess we could set some more defaults in case someone is still targeting netstandard2.0, but there are other rough edges there so I think we'll rather just move folks in the right direction and promote using the newer TFM's. I think it has been in our README for a while now as well, so this will most likely hurt some early adopters so that's why I'm calling it out here. @ErikEJ Or do you have a different opinion on this?

baronfel: This is happening because you're allowing the SDK Container Targets to choose your base image (in the [`ComputeContainerBaseImage` Target](https://github.com/dotnet/sdk/blob/main/src/Containers/packaging/build/Microsoft.NET.Build.Containers.targets#L36-L70)). You shouldn't do this - this Target knows nothing about your scenario or requirements, it is specifically geared for managed .NET applications. The expectation is that projects with different execution requirements create a Target of their own with `BeforeTargets="ComputeContainerBaseImage"` and compute their own value for the `ContainerBaseImage` property, as well as any other data you want to add to the generated containers:
* ContainerLabel Items for key/value metadata
* ContainerEnvironmentVariables Items for, well, container run-time environment variable settings :)
* ContainerAppCommand/ContainerAppCommandArgs to control what the entrypoint/default execution behavior of the container should be

[This](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.Publish.targets#L24-L48) is how the Azure Functions team integrated into container publish, if it helps to see a worked example. Azure Functions projects have separate base images, commands, and require specific env vars in order for the Azure Functions hosts to work. I see a similar need to customize the experience here.

If you integrate in this way, then you can either

* use the TFM as part of your own decision-tree for which base image(s) to use, or
* ignore it completely if there's some good-enough base image that works for all of your use cases.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -310,4 +312,16 @@
...
   </Target>
+
+  <!-- Prepares the publish folder with the necessary files for the resulting container -->
+  <Target Name="PrepareContainer" DependsOnTargets="CopyDacpacs">
+    <PropertyGroup>
+      <TempStagingFolder>$([System.IO.Path]::GetTempPath())</TempStagingFolder>
+      <SqlPackageStagingFolder>$([System.IO.Path]::Combine('$(TempStagingFolder)', 'sqlpackage-linux'))</SqlPackageStagingFolder>
+      <SqlPackageExtractPath>$([System.IO.Path]::Combine('$(PublishDir)', 'sqlpackage'))</SqlPackageExtractPath>
+    </PropertyGroup>
+    <DownloadFile SourceUrl="https://aka.ms/sqlpackage-linux" DestinationFolder="$(TempStagingFolder)" Condition="!Exists('$(SqlPackageExtractPath)')" />
```
nmummau: I personally don't love the idea that we will "always points to the latest release". Can we give the option to point to a specific version of sqlpackage?

For me I personally feel this way just so that as I use the MSBuild.Sdk.SqlProj, I'm in full control of what's happening. If we will "always point to the latest release", that means if sqlpackage ever introduces a breaking change, that might cause some pain. Also the current latest version I just downloaded was `sqlpackage-linux-x64-en-170.2.70.1.zip`, and if you look at Microsoft's own [sqlpackage release notes page here](https://learn.microsoft.com/en-us/sql/tools/sqlpackage/release-notes-sqlpackage?view=sql-server-ver17) they don't even have a release note written for that version yet. 

All of this to me means it would be nice if I were able to optionally pin to a certain version and maintain that dependency myself.

nmummau: Also, if I'm using a pinned version of sqlpackage in other parts of my solution, I might prefer that I use the same sqlpackage version be used across the board. The fact that this SDK uses latest version and some other part of my solution might use a different version introduces stability/consistency concerns.

I just feel that fundamentally it would be good if those using this feature could be in control of the version.

dzsquared: @nmummau - point of clarification - 170.2.70.1 is 170.2.70, we don't include the built number in the rel notes. It matches on https://www.nuget.org/packages/Microsoft.SqlPackage#versions-body-tab

apologies that the download folder name creates confusion, will look into that in the future.

nmummau: @dzsquared Thanks for that. Would be nice for the release notes page to match the downloaded version to remove ambiguity.

nmummau:
@ErikEJ  one more thing I wanted to mention to you that this feature unlocks. For EFCPT, this will be sweet for reverse engineering! If this new publish feature is available, you can now deploy a database project to a containerized database, then run EFCPT reverse engineering against that containerized database, and map the models that are generated back to the host machine that is running the container. It's really amazing. This also removes the need to have support to be able to reverse engineer a dacpac since you are able to reverse engineer against the containerized database directly!

I have a working example of what I described above, but it becomes MUCH easier once the sqlpackage publish is built into this SDK.

README.md
ErikEJ: We need to stick to .NET 8 or higher due to our SQL analyzer package support.

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
ErikEJ: We could make that optional later based on feedback, but right now we get the latest released version from a short url, and I never expect breaking changes from DacFX.

README.md
```diff
@@ -485,13 +485,20 @@ You can now reference your dacpac as a `PackageReference`!
...
 
 ## Publishing support
+Starting with MSBuild.Sdk.SqlProj version 4.0.0 there are two modes of publishing supported: publishing the database directly from the project to a SQL Server instance, or publishing a container image that includes both SqlPackage and the .dacpac ready to be run anywhere a container can be executed.
 
-There is support for publishing a project to a SQL Server using the `dotnet publish` command. This support is designed to be used by developers to deploy or update their local development database quickly. For more advanced deployment scenario's we suggest using [SqlPackage](https://docs.microsoft.com/en-us/sql/tools/sqlpackage/sqlpackage?view=sql-server-ver15) instead as it provides more options.
+> Note: To support both modes we unfortunately had to make a breaking change to the SDK. If you've previously used `dotnet publish` to deploy your database directly to SQL Server you'll now need to add `/t:PublishDatabase` to your command line to retain the previous behavior.
```
jeffrosenberg: I might recommend specifying the version here, i.e. "Note for 4.0.0: ..."

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
jeffrosenberg: > I never expect breaking changes from DacFX
Yeah, I agree with this. Personally, I don't feel too strongly here. I can't imagine it would be much work to default to the latest but allow people to override; on the other hand I have a hard time imagining there generally being an advantage to using an older version.

jmezach: I've had a look at supporting different versions of SqlPackage, but the main issue with that is figuring out the URL to download it from. SqlPackage releases aren't published on GitHub and the `download.microsoft.com` URL's seem to have a GUID that I don't know how to figure out. So unless there's a file somewhere that ties a release version to a download URL I don't think there's a very clean way of supporting this.

That being said we might want to put a `$(SqlPackageDownloadUrl)` property in Sdk.props that points to the "latest" download URL so that folks could set a different value in their `. csproj` to override that if they want.

README.md
jmezach: Fair enough, added that now

README.md
jmezach: @baronfel Thanks for that pointer. I've adopted that approach now and it now works with netstandard2.0 targeting projects as well.

jmezach:
@ErikEJ I think we're now in a good enough place for a preview release? Perhaps you wanna give this another review, since I did make some changes ;).

PR #868, opened 2026-03-17
Create ER diagram during build

### LINKED ISSUE #867: Create Mermaid ER diagram as part of build
I have some code present to do this already, just need to rewrite to use DacFX to enumerate objects.

I think this would be a nice fringe benefit of using this SDK!

Image

@jmezach Thoughts?

ErikEJ:
fixes #867 

@jmezach Decided to make it optional, and add to templates to make it visible.


src/DacpacTool/Diagram/DatabaseModelToMermaid.cs
```diff
@@ -0,0 +1,124 @@
...
+            Lookup['_'] = true;
+            Lookup['-'] = true;
+        }
+
+        public DatabaseModelToMermaid(IReadOnlyList<SimpleTable> tables)
+        {
+            this.tables = tables;
+        }
+
+        public string CreateMermaid(bool createMarkdown = true)
```
jmezach: `createMarkdown` will now always default to true. Do we want to make this configurable?

src/DacpacTool/Diagram/DatabaseModelToMermaid.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using MSBuild.Sdk.SqlProj.DacpacTool.Diagram.Model;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool.Diagram
+{
+    public class DatabaseModelToMermaid
+    {
+        private static readonly bool[] Lookup = new bool[65536];
```
jmezach: I think this is allocating 64 KB of memory which feels like a bit to much for what we're using it for. Perhaps we should look at `BitVector32` or `BitArray`?

src/DacpacTool/Diagram/DacpacModelFactory.cs
```diff
@@ -0,0 +1,330 @@
...
+                {
+                    Name = name,
+                    Schema = schema,
+                    Comment = tableComment,
+                };
+
+                foreach (var colObj in tableObj.GetReferenced(Table.Columns, DacQueryScopes.UserDefined))
+                {
+                    var colType = colObj.GetMetadata<ColumnType>(Column.ColumnType);
+                    if (colType == ColumnType.ComputedColumn)
```
jmezach: Do we not want computed columns in the output? Or should that be a configurable choice maybe?

test/DacpacTool.Tests/DiagramBuilderTests.cs
```diff
@@ -0,0 +1,53 @@
...
+            testConsole.Lines.Count.ShouldBe(1);
+
+            testConsole.Lines.ShouldContain($"Generating ER diagram MyPackage_erdiagram.md");
+
+            var diagramText = File.ReadAllText(diagram, Encoding.UTF8);
+
+            diagramText.ShouldContain("erDiagram");
+
+            diagramText.ShouldContain("    Column1 nvarchar(100) PK");
+        }
```
jmezach, line 45: Perhaps we want to assert here that the stored proc is not part of the diagram?

src/DacpacTool/Diagram/DatabaseModelToMermaid.cs
ErikEJ: Let's keep it as markdown, this was a point in time solution for older VS versions

src/DacpacTool/Diagram/DacpacModelFactory.cs
jeffrosenberg: I don't understand why we wouldn't want them, seems to me they're still part of the table definition

src/DacpacTool/Diagram/Model/SimpleTable.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool.Diagram.Model
+{
+    public class SimpleTable
+    {
+        public string Name { get; set; } = string.Empty;
+
+        public string Schema { get; set; } = string.Empty;
+
+        public string? Comment { get; set; }
```
jeffrosenberg: I'm interested to see comments as part of this model; I'd like to see what they look like in the generated output

src/DacpacTool/Diagram/DacpacModelFactory.cs
ErikEJ: Yeah, this was ported from use with EF Core, where computed columns were broken - I will bring them back!

src/DacpacTool/Diagram/Model/SimpleTable.cs
ErikEJ: This was originally used with EF Core, in a diagram columns do not make sense, so will remove

ErikEJ:
@jmezach A smaller example like the one here?? https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/867

README.md
```diff
@@ -602,6 +604,37 @@ The database name for the create script gets resolved in the following manner:
 - the generated script also uses the resolved database name via a setvar command.
 - if `IncludeCompositeObjects` is true, the composite objects (tables, etc.) from external references are also included in the generated script. This property defaults to `true`
 
+## E-R diagram
```
nmummau: I like the feature overall. I do have one naming/docs suggestion though:

Would we consider GenerateEntityRelationshipDiagram instead of GenerateErDiagram?

ER is a familiar acronym in the database space, but Er looks a little awkward in a public MSBuild property name. Since this is part of the user-facing config surface, I think optimizing for readability is worth the extra length.

Also, in the docs, I don’t think E-R diagram is the right form. The common usage is either entity-relationship diagram, ER diagram, or ERD. The hyphenated acronym looks unusual.

Using the full term at least once in the docs would also help discoverability. People are much more likely to search for “entity relationship diagram” or “entity-relationship diagram” than for just “ER diagram” or "E-R diagram", so having the expanded phrase in the heading/body would make the feature easier to find.

So my preference would be:

property: GenerateEntityRelationshipDiagram
docs: use entity-relationship (ER) diagram on first mention, then ER diagram afterward

README.md
```diff
@@ -602,6 +604,37 @@ The database name for the create script gets resolved in the following manner:
...
+
+The SDK supports generating an E-R diagram from your project. To enable this, add the `GenerateErDiagram` property to your project file:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/4.0.2">
+  <PropertyGroup>
+    <GenerateErDiagram>True</GenerateErDiagram>
```
nmummau: Again love this feature, and I commented on the issue too. Could the SDK provide a way to specify the target destination of where the file ends up when it's generated? Or could the docs elaborate on how you could easily move that file as a post-step after it is generated? I see it being very useful to send this new .md file to a specific spot in my repository. 

It seems that current implementation would save the TestProject_erdiagram.md to the build output directory, is that right?

ErikEJ: @nmummau Correct, but maybe the project root is a better default location? What do you think, @jmezach @jeffrosenberg ??

nmummau: I think the project root is fine to start. Once I use it, I'll offer suggestion if it's awkward.

ErikEJ: Let's hear the verdict of @jmezach - agree the project root may be more useful

README.md
ErikEJ: Implemented, thanks for the suggestion

README.md
nmummau: I like project root, so it can be referenced in docs, or can be gitignored if desired.

jmezach: As mentioned in #867 I think it does make sense to output it to the project root by default

jmezach (approved):
LGTM. I think it makes sense to ship this as a 4.1 release. Just some merge conflicts to resolve now though ;).

src/DacpacTool/Diagram/Model/SimpleColumn.cs
```diff
@@ -0,0 +1,13 @@
+﻿#nullable enable
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool.Diagram.Model
+{
+    public class SimpleColumn
+    {
+        public string Name { get; set; } = string.Empty;
+
+        public string? StoreType { get; set; }
```
nmummau, line 9: Doing a little playing around with Codex on my machine as I look over your changes:

I think the code that creates SimpleColumn instances already always assigns StoreType. The only production path that constructs these diagram model objects already guarantees StoreType is populated. The nullable property type is looser than it needs to be.

  Specifically in src/DacpacTool/Diagram/DacpacModelFactory.cs:23:

  - each column gets storeType from src/DacpacTool/Diagram/
    DacpacModelFactory.cs:202
  - computed columns overwrite that with computed(...) at src/DacpacTool/
    Diagram/DacpacModelFactory.cs:67
  - even the fallback path returns "unknown" at src/DacpacTool/Diagram/
    DacpacModelFactory.cs:207
  - then SimpleColumn.StoreType is assigned unconditionally at src/DacpacTool/
    Diagram/DacpacModelFactory.cs:73
 
```suggestion
        public string StoreType { get; set; } = string.Empty;
```

jmezach:
@ErikEJ Looks like there are some conflicts here now that need resolving. I would love for this feature to land soon, perhaps @nmummau can do a review on this?

ErikEJ:
@jmezach Merge conflict fixed now

src/DacpacTool/Diagram/Model/SimpleTable.cs
```diff
@@ -0,0 +1,19 @@
+﻿#nullable enable
```
jmezach, line 1: Haven't we enabled this at the project level already?

src/DacpacTool/Diagram/DatabaseModelToMermaid.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public DatabaseModelToMermaid(IReadOnlyList<SimpleTable> tables)
+        {
+            this.tables = tables;
+        }
+
+        private static bool IsValidChar(char c) =>
+            c is (>= '0' and <= '9') or (>= 'A' and <= 'Z') or (>= 'a' and <= 'z') or '.' or '_' or '-';
+
+        public string CreateMermaid()
```
jmezach, line 21: I'm wondering if it might be worthwhile to split this method up into smaller parts, one for handling tables, one for handling columns and one for foreign keys. Might make it bit more maintainable.

jmezach (approved):

src/DacpacTool/Diagram/Model/SimpleTable.cs
ErikEJ: No, only warnings

nmummau (approved):

jmezach:
@ErikEJ I would suggest doing an override here so that we can merge this. My guess is that @jeffrosenberg doesn't currently have the bandwidth to review this (which is totally fine) and we already have two reviewers + copilot feedback anyway.

ErikEJ:
@jmezach sounds good, but only you can override 😳

jmezach:
@ErikEJ Done. Shall we ship this as 4.1?

ErikEJ:
Let's do that!

PR #870, opened 2026-03-21
performance refactors

nmummau:
My motivation for this pull request was to see if there was any low-hanging fruit changes that could be made to for performance gains. 

#### refactor: clarify analyzer override tests
  I added tests around rule override behavior before changing code so the behavior contract is explicit and protected. The key gaps were in case-insensitive wildcard matching for both `+!` overrides and `-` suppressions. Those cases were previously untested in this repo path, so I added focused coverage.

#### perf(analyzer): reduce rule parsing overhead
  I changed rule parsing to be single-pass and data-shape friendly:
  - `BuildRuleLists` now parses once instead of doing multiple filtered `Split` loops.
  - `+!` rules are normalized up front into:
    - exact rule IDs (`+!SRD0006`), and
    - wildcard prefixes (`+!SRD*` stored as `SRD`).
  - `SqlRuleProblemExtensions.GetOutputMessage` now checks those pre-normalized prefixes directly
  while formatting, instead of slicing wildcard patterns repeatedly during output formatting.

#### perf(analyzer): speed up suppressor checks
I replaced per-problem linear suppression scans with an index:
  - Build a dictionary once per analysis run:
    - `Dictionary<RuleId, HashSet<SourceName>>`
  - Normalize and store suppression source paths once when loading suppressions.
  - In `SetProblemSuppressor`, suppress-check is now:
    - rule-id lookup → set membership check,
    instead of comparing each diagnostic against every suppression entry.

## why this helps:
  - Before: each diagnostic did `#suppressions` comparisons (O(diag_count × suppress_count)).
  - After: each diagnostic does near-constant work (lookup + contains), plus one-time index build.

  Concrete example:
  - 2,000 diagnostics, 500 suppressions
    - before: ~1,000,000 checks
    - after: ~2,000 lookups + 500 index inserts

  Validation:
  - Added/extended assertions around suppression handling (`RunsAnalyzerWithSuppressionFile` now
  verifies suppressor processing is exercised).
  - Focused analyzer test sets pass on `net8`, `net9`, `net10`.

  Net effect:
  - No behavior changes to outputs.
  - Safer scaling for large projects and suppression-heavy workloads.
  - Tests now lock in the exact behavior I optimized.

### DIFF
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -203,6 +203,11 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
 
         private static string ParseExternalParts(string externalParts)
         {
+            return ParseExternalParts(externalParts, ExternalPartsRegex);
+        }
+
+        private static string ParseExternalParts(string externalParts, Regex externalPartsRegex)
+        {
             string serverVariableName = null;
             string databaseVariableName = null;
             string databaseVariableLiteralValue = null;
@@ -210,21 +215,35 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             // If there are '=' sign in argument assumes that this is formula, else assume that a single value passed and that it is database literal.
             if (externalParts.Contains('=', StringComparison.Ordinal))
             {
-                foreach (Match match in ExternalPartsRegex.Matches(externalParts))
+                try
                 {
-                    if (match.Groups["dbl"].Success)
+                    foreach (Match match in externalPartsRegex.Matches(externalParts))
                     {
-                        databaseVariableLiteralValue = Identifier.EncodeIdentifier(match.Groups["dbl"].Value);
+                        if (match.Groups["dbl"].Success)
+                        {
+                            databaseVariableLiteralValue = Identifier.EncodeIdentifier(match.Groups["dbl"].Value);
+                        }
+                        else if (match.Groups["dbv"].Success)
+                        {
+                            databaseVariableName =
+                                Identifier.EncodeIdentifier(EnsureIsDelimited(match.Groups["dbv"].Value));
+                        }
+                        else if (match.Groups["srv"].Success)
+                        {
+                            serverVariableName = Identifier.EncodeIdentifier(EnsureIsDelimited(match.Groups["srv"].Value));
+                        }
                     }
-                    else if (match.Groups["dbv"].Success)
-                    {
-                        databaseVariableName =
-                            Identifier.EncodeIdentifier(EnsureIsDelimited(match.Groups["dbv"].Value));
-                    }
-                    else if (match.Groups["srv"].Success)
-                    {
-                        serverVariableName = Identifier.EncodeIdentifier(EnsureIsDelimited(match.Groups["srv"].Value));
-                    }
+                }
+                catch (RegexMatchTimeoutException ex)
+                {
+                    throw new ArgumentException(
+                        "Unable to parse reference external parts. " +
+                        "Use a database literal or SQLCMD variable metadata such as " +
+                        "'DatabaseVariableLiteralValue=\"MyDatabase\"', " +
+                        "'DatabaseSqlCmdVariable=\"MyDatabaseVar\"', or " +
+                        "'DatabaseSqlCmdVariable=\"MyDatabaseVar\" ServerSqlCmdVariable=\"MyServerVar\"'.",
+                        nameof(externalParts),
+                        ex);
                 }
             }
             else
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -19,6 +19,8 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
         static Type CustomSchemaDataType;
 
         static MethodInfo SetMetadataMethod;
+        private static readonly Regex ExternalPartsRegex = new Regex(@"dbl=(?<dbl>\w+)|dbv=(?<dbv>\w+)|srv=(?<srv>\w+)", 
+            RegexOptions.CultureInvariant | RegexOptions.Compiled, TimeSpan.FromSeconds(1));
 
 #pragma warning disable CA1810 // Initialize reference type static fields inline
         static Extensions()
@@ -206,10 +208,9 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
             string databaseVariableLiteralValue = null;
 
             // If there are '=' sign in argument assumes that this is formula, else assume that a single value passed and that it is database literal.
-            if (externalParts.Contains('=', StringComparison.OrdinalIgnoreCase))
+            if (externalParts.Contains('=', StringComparison.Ordinal))
             {
-                foreach (Match match in new Regex(@"dbl=(?<dbl>\w+)|dbv=(?<dbv>\w+)|srv=(?<srv>\w+)",
-                    RegexOptions.CultureInvariant, TimeSpan.FromSeconds(1)).Matches(externalParts))
+                foreach (Match match in ExternalPartsRegex.Matches(externalParts))
                 {
                     if (match.Groups["dbl"].Success)
                     {
@@ -328,17 +329,16 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 throw new InvalidOperationException("Unable to create instance of CustomSchemaData.");
             }
 
+            var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
+
+            if (setMetadataMethod == null)
+            {
+                throw new InvalidOperationException("Unable to find SetMetadata method on CustomSchemaData.");
+            }
+
             foreach (var variableName in variables)
             {
                 Console.WriteLine($"Adding SqlCmd variable {variableName}");
-
-                var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
-
-                if (setMetadataMethod == null)
-                {
-                    throw new InvalidOperationException("Unable to find SetMetadata method on CustomSchemaData.");
-                }
-
                 setMetadataMethod.Invoke(customData, new object[] { variableName, string.Empty });
             }
 
--- a/src/DacpacTool/Extensions.cs
+++ b/src/DacpacTool/Extensions.cs
@@ -329,17 +329,10 @@ namespace MSBuild.Sdk.SqlProj.DacpacTool
                 throw new InvalidOperationException("Unable to create instance of CustomSchemaData.");
             }
 
-            var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
-
-            if (setMetadataMethod == null)
-            {
-                throw new InvalidOperationException("Unable to find SetMetadata method on CustomSchemaData.");
-            }
-
             foreach (var variableName in variables)
             {
                 Console.WriteLine($"Adding SqlCmd variable {variableName}");
-                setMetadataMethod.Invoke(customData, new object[] { variableName, string.Empty });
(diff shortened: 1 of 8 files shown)


src/DacpacTool/DacpacTool.csproj
```diff
@@ -4,14 +4,25 @@
...
     <TargetFrameworks>net8.0;net9.0;net10.0</TargetFrameworks>
-    <LangVersion>9.0</LangVersion>
     <AnalysisMode>all</AnalysisMode>
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <AnalysisLevel>latest</AnalysisLevel>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <Nullable>warnings</Nullable>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
ErikEJ: Remove an let the TFM above decide

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

test/DacpacTool.Tests/DacpacTool.Tests.csproj
```diff
@@ -5,11 +5,22 @@
     <IsPackable>false</IsPackable>
     <AssemblyName>MSBuild.Sdk.SqlProj.DacpacTool.Tests</AssemblyName>
     <RootNamespace>MSBuild.Sdk.SqlProj.DacpacTool.Tests</RootNamespace>
-    <LangVersion>9.0</LangVersion>
     <NuGetAudit>false</NuGetAudit>
     <TestTfmsInParallel>false</TestTfmsInParallel>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
ErikEJ: Let the TFM decide by removing Langversion

test/DacpacTool.Tests/PackageDeployerTests.cs
```diff
@@ -189,6 +189,28 @@ public void SetPropertySimpleValue()
             packageDeployer.DeployOptions.AllowDropBlockingAssemblies.ShouldBeTrue();
         }
 
+        [TestMethod]
+        public void SetPropertyCaseInsensitiveNames()
```
ErikEJ, line 193: Wonder if the correct behaviour is case sensitive?

nmummau: Possibly. However, this test matches pre-existing behavior. Before the refactor, property names were already matched with StringComparer.OrdinalIgnoreCase; the new dictionary-based lookup keeps that same case-insensitive behavior and just avoids repeated reflection.

I wrote this test to prove that the behavior of the refactor wouldn't introduce a regression.

The old code was already case-insensitive here in PropertyParser.cs
```csharp
.SingleOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase));
```
The new code preserves that same behavior by moving the lookup into a dictionary created with StringComparer.OrdinalIgnoreCase:
```csharp
new Dictionary<string, PropertyInfo>(StringComparer.OrdinalIgnoreCase)
```

src/DacpacTool/PackageAnalyzer.cs
```diff
@@ -53,28 +56,36 @@ public void Analyze(TSqlModel model, FileInfo outputFile, FileInfo[] analyzers)
...
                     var problemSuppressor = ProjectProblemSuppressor.CreateSuppressor(projectDir);
-
-                    suppressedProblems = problemSuppressor.GetSuppressedProblems().ToList();
+                    var suppressedProblems = problemSuppressor.GetSuppressedProblems();
 
                     foreach (var problem in suppressedProblems)
                     {
                         _console.WriteLine($"Suppressing rule: '{problem.Rule.RuleId}' in '{problem.SourceName}'");
+                        var sourceName = Path.Combine(projectDir, problem.SourceName)
+                            .Replace('\\', Path.AltDirectorySeparatorChar);
```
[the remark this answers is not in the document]
nmummau: This logging was already in place before my changes
```charp
_console.WriteLine($"Suppressing rule: '{problem.Rule.RuleId}' in '{problem.SourceName}'");
```

So I'm pretty sure Copilot is confused.

src/DacpacTool/SqlRuleProblemExtensions.cs
```diff
@@ -22,11 +23,15 @@ public static string GetOutputMessage(this SqlRuleProblem sqlRuleProblem, HashSe
                 sqlRuleProblemSeverity = SqlRuleProblemSeverity.Error;
             }
 
-            var wildCardErrorRules = errorRules
-                .Where(r => r.EndsWith('*'));
-            if (wildCardErrorRules.Any(s => sqlRuleProblem.RuleId.StartsWith(s[..^1], StringComparison.OrdinalIgnoreCase)))
+            // Wildcard severity overrides are stored as trimmed prefixes (for "+!SRD*",
+            // we store "SRD"), so we only keep this loop for prefix rules.
```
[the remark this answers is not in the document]
nmummau: Valid. Fixed in a new commit: https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/870/changes/56d0e9b5b4c85c4e2abdbc34765af0b8542fb6ba

src/DacpacTool/Extensions.cs
```diff
@@ -328,17 +329,16 @@ public static void AddSqlCmdVariables(this TSqlModel model, string[] variables)
                 throw new InvalidOperationException("Unable to create instance of CustomSchemaData.");
             }
 
+            var setMetadataMethod = customData.GetType().GetMethod("SetMetadata", BindingFlags.Public | BindingFlags.Instance);
+
+            if (setMetadataMethod == null)
+            {
+                throw new InvalidOperationException("Unable to find SetMetadata method on CustomSchemaData.");
+            }
```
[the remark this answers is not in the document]
nmummau: Valid. Fixed in https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/870/changes/a94f97146c0a2372ccad21c31ba6207fc3416936

src/DacpacTool/PackageAnalyzer.cs
```diff
@@ -13,6 +13,8 @@ public sealed class PackageAnalyzer
         private readonly HashSet<string> _ignoredRules = new();
         private readonly HashSet<string> _ignoredRuleSets = new();
         private readonly HashSet<string> _errorRuleSets = new();
+        private readonly List<string> _errorRulePrefixes = new();
+        private readonly Dictionary<string, HashSet<string>> _suppressedProblemsByRule = new(StringComparer.Ordinal);
```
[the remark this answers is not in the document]
nmummau: Valid. Fixed here: https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/870/changes/6edf6e2e5d1fb5e0145667a24fd393416c7edf9f

src/DacpacTool/Extensions.cs
```diff
@@ -19,6 +19,8 @@ public static class Extensions
         static Type CustomSchemaDataType;
 
         static MethodInfo SetMetadataMethod;
+        private static readonly Regex ExternalPartsRegex = new Regex(@"dbl=(?<dbl>\w+)|dbv=(?<dbv>\w+)|srv=(?<srv>\w+)", 
+            RegexOptions.CultureInvariant | RegexOptions.Compiled, TimeSpan.FromSeconds(1));
```
jmezach, line 23: What happens when the regex times out?

ErikEJ: Having a regex timeout is always recommended as a security measure against DOS attacks

src/DacpacTool/PropertyParser.cs
```diff
@@ -43,10 +49,12 @@ public static DacDeployOptions ExtractDeployOptions(this BuildOptions options)
 
         private static PropertyInfo GetDacDeployOptionsProperty(string propertyName)
         {
-            var property = typeof(DacDeployOptions).GetProperties()
-                .SingleOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase));
+            if (DacDeployOptionProperties.TryGetValue(propertyName, out var property))
```
jmezach: Does this change actually improve performance? From what I understand of the change we're now creating a dictionary of property names to their associated `PropertyInfo` objects once at startup of the CLI tool and then using that when a property is set. But most projects will probably only ever specify a few properties and we'd only be looking up those and we'd be only looking those up once per invocation of the tool, so I don't think really helps much and only has the CLI tool use more memory for no apparent gain.

src/DacpacTool/Extensions.cs
jmezach: Sure, but we're not handling the exception that I'm guessing is being thrown. So if it times out the user would get a weird exception with a stack trace that probably won't make much sense

src/DacpacTool/PropertyParser.cs
nmummau: You’re right that this is a micro-optimization. The cache avoids repeated reflection and linear scans, but SetDeployProperty is only called for the few properties a project specifies, so the practical gain is likely negligible. I’m happy to revert it and keep the simpler implementation.

nmummau: removed it here: https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/870/changes/e4e27d705a24f30c2b05b4e4a7bccc5d91337cd9

src/DacpacTool/Extensions.cs
nmummau: Fair point. Fixed here with a try/catch and a new test! https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/870/changes/c5aa921ab3070002c27b5f9d40eaee702def2a92

src/DacpacTool/PackageAnalyzer.cs
```diff
@@ -13,6 +13,8 @@ public sealed class PackageAnalyzer
         private readonly HashSet<string> _ignoredRules = new();
         private readonly HashSet<string> _ignoredRuleSets = new();
```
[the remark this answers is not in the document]
nmummau: fixed

src/DacpacTool/Extensions.cs
```diff
@@ -200,31 +202,49 @@ public static void AddReference(this TSqlModel model, string referencePath, stri
...
+
+        private static string ParseExternalParts(string externalParts, Regex externalPartsRegex)
         {
             string serverVariableName = null;
             string databaseVariableName = null;
             string databaseVariableLiteralValue = null;
 
             // If there are '=' sign in argument assumes that this is formula, else assume that a single value passed and that it is database literal.
-            if (externalParts.Contains('=', StringComparison.OrdinalIgnoreCase))
+            if (externalParts.Contains('=', StringComparison.Ordinal))
```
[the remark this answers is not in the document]
nmummau: Copilot is hallucinating. `if (externalParts.Contains('=', StringComparison.Ordinal))` is totally valid in net8/net9

src/DacpacTool/SqlRuleProblemExtensions.cs
```diff
@@ -22,11 +23,16 @@ public static string GetOutputMessage(this SqlRuleProblem sqlRuleProblem, HashSe
                 sqlRuleProblemSeverity = SqlRuleProblemSeverity.Error;
             }
```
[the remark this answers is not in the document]
nmummau: In src/DacpacTool/PackageAnalyzer.cs:15, _errorRuleSets is already initialized with StringComparer.OrdinalIgnoreCase

jmezach (approved):
LGTM

ErikEJ (approved):

ErikEJ:
@nmummau Could you look at the build error?

nmummau:
> @nmummau Could you look at the build error?

Fixed here https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/892. This was a result of the CI check we put in place to add `dotnet format style` to remove unused usings

Issue #872, opened 2026-03-21
Feature Request: Item-based CodeAnalysisRule configuration as an alternative to CodeAnalysisRules


nmummau:
## Problem

I really like the static analysis rules provided by `ErikEJ.DacFX.SqlServer.Rules` and `ErikEJ.DacFX.TSQLSmellSCA`. I often use `+!*` in `<CodeAnalysisRules>` so that all analyzer violations are treated as build-breaking errors.

However, this also means I sometimes need to globally suppress rules that are not relevant for a particular project. Today, the only way to do that is by building a long semicolon-delimited string inside `<CodeAnalysisRules>`.

For example:

```xml
<Project Sdk="MSBuild.Sdk.SqlProj/4.0.0">
  <PropertyGroup>
    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
    <!--SQL Code Analysis
        Suppressed rules:
        • SRD0004 (Index on Foreign Key)                https://github.com/ErikEJ/SqlServer.Rules/blob/master/docs/Design/SRD0004.md
        • SRD0010 (Low identity seed value)             https://github.com/ErikEJ/SqlServer.Rules/blob/master/docs/Design/SRD0010.md
        • SRD0047 (Avoid deprecated collation syntax)   https://github.com/ErikEJ/SqlServer.Rules/blob/master/docs/Design/SRD0047.md
        • SRD0058 (Ordinal parameters used)             https://github.com/ErikEJ/SqlServer.Rules/blob/master/docs/Design/SRD0058.md
        • SRD0062 (Use proper collation in temp tables) https://github.com/ErikEJ/SqlServer.Rules/blob/master/docs/Design/SRD0062.md
        • SRN0007 (Naming convention)                   https://github.com/ErikEJ/SqlServer.Rules/blob/master/docs/Naming/SRN0007.md
    -->
    <!-- Disable specific rules, then treat remaining rule violations as errors (+!*) -->
    <CodeAnalysisRules>-SqlServer.Rules.SRD0004;-SqlServer.Rules.SRD0010;-SqlServer.Rules.SRD0047;-SqlServer.Rules.SRD0058;-SqlServer.Rules.SRD0062;-SqlServer.Rules.SRN0007;+!*</CodeAnalysisRules>
  <PropertyGroup>
</Project>
```

## Proposal
Add an item-based alternative for configuring analyzer rule behavior, for example:

```xml
[...]

PR #874, opened 2026-03-21
feat: support item-based CodeAnalysisRule config

nmummau:
For issue #872, introduce `CodeAnalysisRule` items as an additive alternative to the existing `CodeAnalysisRules` property.

The new format improves readability for projects with many analyzer rule overrides, while remaining backward-compatible with the current semicolon-delimited property-based configuration.

Also adds README documentation and targeted tests covering:
- wildcard RuleId handling
- combination with existing CodeAnalysisRules
- +!* global error behavior
- suppression precedence on conflicts

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,60 @@ A xml file with the analysis results is created in the output folder.
 
 The optional `CodeAnalysisRules` property allows you to disable individual rules or groups of rules for the entire project.
 
+If you have many analyzer rule adjustments, you can also declare them as items instead of a single semicolon-delimited property value:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/4.0.0">
+  <PropertyGroup>
+    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CodeAnalysisRule Include="_">
+      <RuleId>SqlServer.Rules.SRD*</RuleId>
+      <Action>Suppress</Action>
+      <Explanation>suppress globally for some good reason</Explanation> <!-- optional explanation -->
+    </CodeAnalysisRule>
+    <CodeAnalysisRule Include="_">
+      <RuleId>SqlServer.Rules.SRN0005</RuleId>
+      <Action>Error</Action>
+      <Explanation>this type of violoation should break the build</Explanation> <!-- optional explanation -->
+    </CodeAnalysisRule>
+  </ItemGroup>
+</Project>
+```
+
+`Include` is only a placeholder required by MSBuild. The SDK uses `RuleId` and `Action`. `Explanation` is optional and is ignored by the SDK, so you can use it as a maintainer note in the project file.
+
+Supported `Action` values are:
+
+- `Suppress` to disable the specified rule
+- `Error` to report the specified rule as an error
+
+Use the `RuleId` element to define the rule name or wildcard pattern. For example, `SqlServer.Rules.SRD*` matches all rules with that prefix.
+
+If both `CodeAnalysisRules` and `CodeAnalysisRule` items are used, they are combined.
+
+You can use `+!*` in `CodeAnalysisRules` to treat all analyzer findings as errors, then add `CodeAnalysisRule` items to suppress specific rules:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/4.0.0">
+  <PropertyGroup>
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -626,6 +626,60 @@ A xml file with the analysis results is created in the output folder.
...
+If you have many analyzer rule adjustments, you can also declare them as items instead of a single semicolon-delimited property value:
+
+```xml
+<Project Sdk="MSBuild.Sdk.SqlProj/4.0.0">
+  <PropertyGroup>
+    <RunSqlCodeAnalysis>True</RunSqlCodeAnalysis>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <CodeAnalysisRule Include="_">
```
jmezach, line 638: I think we should have the value of `RuleId` in the `Include` attribute instead. That would be more idiomatic MSBuild. We can access that value using `%(CodeAnalysisRule.Identity)`, see https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-well-known-item-metadata?view=visualstudio

README.md
```diff
@@ -626,6 +626,60 @@ A xml file with the analysis results is created in the output folder.
...
+Supported `Action` values are:
+
+- `Suppress` to disable the specified rule
+- `Error` to report the specified rule as an error
+
+Use the `RuleId` element to define the rule name or wildcard pattern. For example, `SqlServer.Rules.SRD*` matches all rules with that prefix.
+
+If both `CodeAnalysisRules` and `CodeAnalysisRule` items are used, they are combined.
+
+You can use `+!*` in `CodeAnalysisRules` to treat all analyzer findings as errors, then add `CodeAnalysisRule` items to suppress specific rules:
```
jmezach, line 663: Couldn't we just use `TreatWarningsAsErrors`?

src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets
```diff
@@ -179,10 +179,41 @@
     <Message Importance="low" Text="Included Files: $(IncludedFiles)" />
   </Target>
 
+  <!--
+    Builds the effective code analysis rule expression from the CodeAnalysisRules property and any CodeAnalysisRule items.
+  -->
+  <Target Name="ComputeCodeAnalysisRules">
+    <ItemGroup>
+      <_CodeAnalysisRuleExpressions Include="@(CodeAnalysisRule)"
```
jmezach, line 187: Can we add some comments here to explain what each of these elements do? We probably understand it now, but do we understand it a year from now

jmezach (changes requested):
Off to a good start, just a couple of comments. Thank you for your contribution @nmummau!

ErikEJ:
@nmummau @jmezach There is a NuGet package out there to do it, used bu for example SQL Formatter from Mads Kristensen

nmummau:
> @nmummau Thoughts on .editorconfig vs this? I think many use .editorconfig for exactly this use case.

That's a much better idea. Something like:
```
dotnet_diagnostic.SRD0006.severity = error
dotnet_diagnostic.SRD0002.severity = none
```

I've never coded anything with analyzers so I'll have to do some digging to see how to implement it that way. Any tips?

ErikEJ:
@nmummau Yeah, as I wrote above, have a look at the SQL Formatter repo, which uses the editorconfig .net library

nmummau:
> @nmummau Yeah, as I wrote above, have a look at the SQL Formatter repo, which uses the editorconfig .net library

Checked it over. [SqlFormatter](https://github.com/madskristensen/SqlFormatter) seems to be using a set of custom keys and ships them in [sqlformatter-editorconfig-schema.json](https://github.com/madskristensen/SqlFormatter/blob/master/src/Schemas/sqlformatter-editorconfig-schema.json)

I think we would need to implement it quite differently as we need to be dynamic and support any fully qualified rules that might be in place.

I'm going to create a separate PR for an implementation I came up with that generically allows for any rules, for example
```
dotnet_diagnostic.SqlServer.Rules.SRD0068.severity = error
dotnet_diagnostic.SqlServer.Rules.SRP0005.severity = none
```

ErikEJ:
@nmummau Looks good, not sure about the "dotnet_diagnostic" prefix though.

nmummau:
> @nmummau Looks good, not sure about the "dotnet_diagnostic" prefix though.

What would you prefer? dotnet_diagnostics seems right to me since it matches the standard severity shape users already recognize [outlined here](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options)

PR #876, opened 2026-03-22
chore(actions): ubuntu-22 -> 24, windows-2022 -> 2025

nmummau:
- Replace ubuntu-22.04 with ubuntu-24.02
- Replace windows-2022 with windows-2025

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
   # Run the build first
   build:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -119,7 +112,7 @@ jobs:
     needs: build
     strategy:
       matrix:
-        os: [ "ubuntu-22.04", "macos-14", "windows-2022" ]
+        os: [ "ubuntu-24.04", "ubuntu-22.04", "macos-14", "windows-2022" ]
```
nmummau: @ErikEJ  @jmezach do we still want to use ubuntu-22 for some reason I'm unaware of? If not, I can update this PR to simply replace 22 with 24 so we don't run so many in parallel. Up to you, but since ubuntu-24 is the LTS, I doubt we need to also run tests on ubuntu-22 unless I'm unaware of a concern you would have with that.

ErikEJ: Drop Ubuntu 22 pls

.github/workflows/main.yml
```diff
@@ -119,7 +112,7 @@ jobs:
     needs: build
     strategy:
       matrix:
-        os: [ "ubuntu-22.04", "macos-14", "windows-2022" ]
+        os: [ "ubuntu-24.04", "macos-14", "windows-2022" ]
```
nmummau: Since we refreshed Ubuntu from 22.04 to 24.04, do we also want to refresh the Windows test runner from windows-2022 to windows-2025, or keep windows-2022 as the baseline?

nmummau: I updated just to try it, and it's all green. If we want to keep the operating systems updated, I think switching both ubuntu and 2025 makes sense. Up to you maintainers though!

PR #878, opened 2026-03-22
feat: quiet SDK progress logging by default

nmummau:
Progress logging from MSBuild.Sdk.SqlProj is now quiet by default during build and publish. Warnings and errors are unchanged.

  This change reduces noisy per-file and per-reference console output on the hot build path while preserving actionable diagnostics. It also fixes a direct console-write path so verbose logging consistently flows through the SDK console abstraction.

  To restore the previous detailed progress output, set: `dotnet build /p:MSBuildSdkSqlProjVerbose=True`

  What changed:

  - Adds a --verbose path inside DacpacTool
  - Gates high-volume progress messages behind verbose output
  - Leaves warnings and errors always visible
  - Wires MSBuildSdkSqlProjVerbose=True through the SDK targets
  - Documents the new property in the README

  Compatibility note:

  - This is not a functional build break, but it is an observable logging/output change
  - Users with CI log scraping or tooling that depends on previous progress lines may need to enable MSBuildSdkSqlProjVerbose=True

  Validation:

  - New logging tests were created and pass on net8.0, net9.0, and net10.0

### DIFF
--- a/test/DacpacTool.Tests/SdkTargetsVerboseTests.cs
+++ b/test/DacpacTool.Tests/SdkTargetsVerboseTests.cs
@@ -1,0 +1,131 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Shouldly;
+
+namespace MSBuild.Sdk.SqlProj.DacpacTool.Tests
+{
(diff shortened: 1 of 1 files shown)


nmummau:
I marked this as a draft because I want to test this at work this coming week.

Issue #788, opened 2025-09-23
SQL generated script is not copied when referenced a MSBuild.Sdk.SqlProj


GillesTourreau:
When we activate the generation of SQL Script with the `<GenerateCreateScript>` property, the generated script is not copied if the database project is referenced by an other project.

To reproduce the issue:
- Create a MSBuild.Sdk.SqlProj database project with one SQL object (a table for example) : MyDatabase.csproj
- Set the following MSBuild properties:

```xml
<Project Sdk="MSBuild.Sdk.SqlProj/3.2.0">
  <PropertyGroup>
      <GenerateCreateScript>True</GenerateCreateScript>
  </PropertyGroup>
</Project>
```
- Create a C# project (unit test for example) : MyUnitTests.csproj
- Add a reference to the `MyDatabase.csproj` project previously created.
- Set the `ReferenceOutputAssembly` property to `false` in the `<ProjectReference>` added.
```
<ItemGroup>
   <ProjectReference Include="..\MyDatabase\MyDatabase.csproj" ReferenceOutputAssembly="False" />
</ItemGroup>
```
- Compile the `MyUnitTests.csproj` project.

The `MyDatabase_Create.sql` is generated in the `bin` output folder of the `MyDatabase.csproj` project. But it is not copied in the `MyUnitTests.csproj` output  `bin` folder.
> NB: For the `MyDatabase.dacpac`, it is generated in the output `bin` folder of the `MyDatabase.csproj` and copied in the `MyUnitTests.csproj` output `bin` folder.

ErikEJ:
That does not look like a documented or supported usage of project references.

ErikEJ:
@jmezach Thoughts?

jmezach:
I agree that this is not a documented feature currently, but I guess we could add the generate script to [this MSBuild target](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/b0d52ef6dbcf203e53983813ded8cb1b7027f550/src/MSBuild.Sdk.SqlProj/Sdk/Sdk.targets#L300) so that it would be copied over to the referenced projects output folder.

nmummau:
@GillesTourreau This is fixed in [v4.1.0 ](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/releases/tag/v4.1.0). Could you let us know how it goes for you?

GillesTourreau:
[...]

PR #879, opened 2026-03-22
fix: copy generated create scripts to referencing project outputs

nmummau:
Include generated *_Create.sql artifacts in the copy-local output items so they flow to referencing project output directories alongside dacpacs.

  Add a regression test covering a project reference with GenerateCreateScript enabled, and document the behavior in the README.

This addresses issue #788


nmummau:
set this PR as a draft cause I want to test on Monday

test/TestIncludeFromVanillaProjWithGeneratedScript/global.json
jmezach, line 1: Do we need this?

test/DacpacTool.Tests/SdkTargetsTests.cs
jmezach, line 1: I'm wondering if this is the best way of testing this. Essentially what we want to do is run `dotnet build` and check if a file is created in the location we expect. This is a lot of code to maintain just to test that. Perhaps a simple script is more appropriate for testing this

test/TestIncludeFromVanillaProjWithGeneratedScript/global.json
nmummau: No, we don't. It's been removed.

test/DacpacTool.Tests/SdkTargetsTests.cs
nmummau: Good point. I'll give this some thought

nmummau: Simplified here https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/879/changes/3234288005e8570e8b67402e2746b9857dbe5001

PR #888, opened 2026-03-27
feat: add .editorconfig support for SQL analyzer rule severity overrides

nmummau:
Per our discussion in PR https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/pull/874 @ErikEJ  suggested to try out an implementation with .editorconfig.

That is what this here PR is for. It is another way to address issue https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/872

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,6 +112,10 @@ jobs:
     - name: dotnet build SDK
       run: dotnet build ./src/MSBuild.Sdk.SqlProj/MSBuild.Sdk.SqlProj.csproj -c Release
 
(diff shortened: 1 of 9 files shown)


ErikEJ:
@nmummau too much code!! Could we use the .editorconfig library?

nmummau:
I think at this point I'll bow out and scrap the effort and continue to deal with with a long line of multiple suppressions living within the single `<CodeAnalysisRules>`

ErikEJ:
@nmummau Fair. If it is any use, VS Code mssql extension now has a UI to manage the rules.

Issue #916, opened 2026-04-16
Support including CLR assemblies


jvdslikke:
We are making use of SQL CLR integration. At the moment there is no easy way to support this in our CI/CD workflows. 

- Installing the assemblies in PostDeployment scripts has the disadvantage of not having the objects in the model. 
- [ErikEJ's workaround](https://erikej.github.io/dacfx/sqlclr/2025/01/28/dacfx-sqlclr-msbuild-sdk-sqlproj.html) has the disadvantage of still having to use Visual Studio and SSDT. 

Can support for CLR assemblies be added? 

I already created a working proof-of-concept: #915

jvdslikke:
PR with new/better/alternative solution: #924

jvdslikke:
It turns out a Dacpac with an assembly can only be created on .NET Framework on Windows. Building with .NET Core gives error `This assembly is corrupt or not valid` in `Microsoft.SqlServer.Dac.DacPackageExtensions.BuildPackage`. 

I created various workarounds: 

* A new .NET Framework DacpacTool which adds the assembly (and all objects referencing the assembly) afterwards
  * #915
  * Disadvantages:
    * Can only build on Windows
    * Has ugly code in `ValidateModel` to detect objects referencing the assembly
    * Adding the objects referencing the assembly (and their references and so on recursive) afterwards can lead to a lot of objects being "deferred"
    * Slower build: needs to load the earlier created Dacpac and save it again

* A new .NET Framework DacpacTool which fully creates the Dacpac when assemblies are referenced. Shared code is moved to a .NET Standard library.  
  * #924
  * Disadvantages:
    * Can only build on Windows
    * Big refactor

* New code in the existing .NET Core DacpacTool which adds the assembly directly into the Dacpac XML
  * #929 
  * Disadvantages:
    * Hacky code: ignores some errors in `ValidateModel` and makes assumptions about the inner structure of the Dacpac XML

I can understand if you don't want to include any of these workarounds.

ErikEJ:
@jvdslikke Thanks for your efforts, this is not something we plan to include - we will point users to the [documented workaround](https://rr-wfm.github.io/MSBuild.Sdk.SqlProj/docs/runtime-and-limitations.html#sql-clr-objects-support)

PR #924, opened 2026-04-24
Include referenced libraries as CLR assembly in Dacpac (full Dacpac creation with .NET Framework)

jvdslikke:
Hi, for #916 I was looking for a way to include another project with SQL CLR code as Assembly in the Dacpac. Because I didn't like to have it in Postdeployment scripts and also didn't want to keep an old SSDT project. I managed to get it working by adding support for DLL references.

The Microsoft Dacpac library validates the assembly which does not work in .NET Core. To work around this I moved all Dacpac creation code and shared code to a .NET Standard library. A new .NET Framework DacpacTool uses this library to create the Dacpac if assemblies are referenced.


ErikEJ:
How is this addition of extra code to maintain  significantly different from having a Windows built .dacpac from a classic .sqlproj?

jvdslikke:
@ErikEJ this extra code enables us to step away from the classic .sqlproj (SSDT) projects and Visual Studio, and fully adopt the new SDK-style projects everywhere. Actually it is not much extra code, some code is moved to a Netstandard library and a very small Framework console app is added that uses it. 

I just discovered that Microsoft.Build.Sql [states in its documentation that it supports CLR](https://learn.microsoft.com/en-us/sql/tools/sql-database-projects/sql-database-projects#original-projects-vs-sdk-style-projects), but I cant get it to work. I created an issue with them: https://github.com/microsoft/DacFx/issues/785

It might even run on Linux with Mono, i'm testing it.

jvdslikke:
> How is this addition of extra code to maintain significantly different from having a Windows built .dacpac from a classic .sqlproj?

@ErikEJ and also we want to use Aspire which does not support classic .sqlproj

jvdslikke:
@ErikEJ apparently DacFX doesn't work on Mono, so yes, Dacpacs with assemblies can only build on Windows

jmezach:
Does referencing an existing `.dacpac` that has a CLR assembly and that was perhaps build with good-old `.sqlproj` and then referencing said `.dacpac` from an MSBuild.Sdk.SqlProj project not work?

ErikEJ:
@jmezach 

Exactly, as described in my sample and blog post.

I do not see an advantage here, just more complex code for us to maintain.

https://erikej.github.io/dacfx/sqlclr/2025/01/28/dacfx-sqlclr-msbuild-sdk-sqlproj.html

jvdslikke:
@jmezach @ErikEJ I want a solution where no manual steps are needed: just change the C# code and rebuild the database project and the C# should be included in the Dacpac. 

Because I don't like this to be a Windows only solution I resorted back to the XML solution and created PR #929 for it. 

I can understand if you don't want to accept any of these PR's. I will summarize my experiments in #916, then let me know you final verdict and I can close the issue and the PRs. I am happy to maintain my own fork until a better solution arrives.

ErikEJ:
@jvdslikke which solution did you end up preferring?

jvdslikke:
@ErikEJ I prefer the "directly into the Dacpac XML" (#916) solution, mostly because it doesn't need Windows. I am not very worried that the Dacpac format changes, I think Microsoft wants to keep it backwards compatible.

ErikEJ:

PR #929, opened 2026-05-04
Include referenced libraries as CLR assembly in Dacpac (afterwards into Dacpac XML)

jvdslikke:
Hi, for https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/issues/916 I was looking for a way to include another project with SQL CLR code as Assembly in the Dacpac. Because I didn't like to have it in Postdeployment scripts and also didn't want to keep an old SSDT project. I managed to get it working by adding support for DLL references.

The Microsoft Dacpac library validates the assembly which does not work in .NET Core. To work around this I add the assembly bytes directly into the Dacpac XML.


ErikEJ:

PR #939, opened 2026-05-21
Emit clear error for invalid enum model properties (#933)

### LINKED ISSUE #933: Model properties using T-SQL syntax (READ_WRITE, AUTO, CHECKSUM) cause unhandled exception in DacpacTool
When setting model properties whose underlying type is a DacFX enum, users naturally reach for the T-SQL syntax that is documented [here](https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/tools/sql-database-projects/concepts/project-properties.md#all-project-properties) (e.g. READ_WRITE, AUTO, CHECKSUM, etc.) — but the SDK forwards the value verbatim to TSqlModelOptions, which expects the C# enum member name (ReadWrite, Auto, Checksum). The result is an unhandled exception inside DacpacTool.dll, exit code -532462766 (0xE0434352), with no indication of which property failed.

  ## Repro

  <PropertyGroup>
    <QueryStoreDesiredState>READ_WRITE</QueryStoreDesiredState>
    <QueryStoreCaptureMode>AUTO</QueryStoreCaptureMode>
  </PropertyGroup>

  Build fails

  ## Expected

  Either:
  1. Accept T-SQL syntax (READ_WRITE → ReadWrite) by normalizing before enum parsing, or
  2. Catch the enum-parse failure and emit a clear error naming the property and listing valid values, e.g.: QueryStoreDesiredState value READ_WRITE is invalid. Expected one of: Off, ReadOnly, ReadWrite.

  ## Workaround

  Use the C# enum member names exactly as defined in Microsoft.SqlServer.Dac.Model.*:
  - QueryStoreDesiredState: Off | ReadOnly | ReadWrite
  - QueryStoreCaptureMode: All | Auto | None
  - PageVerifyMode: Unknown | Checksum | TornPageDetection | None

  ## Environment

  - MSBuild.Sdk.SqlProj 4.2.0
  - DacFX 162.5.57

ErikEJ:
Setting a model property like QueryStoreDesiredState=READ_WRITE used T-SQL syntax that DacFX's TSqlModelOptions doesn't accept, and the resulting Enum.Parse ArgumentException bubbled out of DacpacTool unhandled, killing the process with exit code 0xE0434352 and no hint which property was wrong.

Replace Enum.Parse with a ParseEnum<T> helper that uses TryParse + IsDefined and throws a message naming the property, echoing the bad value, and listing valid C# enum members. Catch ArgumentException around the BuildProperty loop in Program.BuildDacpac so it surfaces as 'ERROR: ...' with exit code 1, matching the DeployDacpac pattern.

fixes #933


src/DacpacTool/Program.cs
```diff
@@ -46,7 +46,15 @@ internal static async Task<int> BuildDacpac(BuildOptions options)
...
                 {
                     string[] keyValuePair = propertyValue.Split('=', 2);
-                    packageBuilder.SetProperty(keyValuePair[0], keyValuePair[1]);
+                    try
+                    {
+                        packageBuilder.SetProperty(keyValuePair[0], keyValuePair[1]);
+                    }
+                    catch (ArgumentException ex)
+                    {
+                        Console.WriteLine($"ERROR: {ex.Message}");
```
jmezach, line 55: Does this show up in the errors window in VS somehow? Would be nice if we could guide the user to the exact location in the .csproj where the invalid value is.

jmezach (approved):

src/DacpacTool/Program.cs
ErikEJ: It shows up in the error window, but no with file and line numbers

jmezach: Hmm, I guess we could emit that, but then we'd have to know from the command line arguments that the exact location is in the .csproj. Maybe it's not worth it, but it would be nice ;).

ErikEJ: True, but this is the first time we have any user issues with this scenario

ErikEJ:
@jeffrosenberg Do you have time to look at this?

jeffrosenberg (approved):

PR #970, opened 2026-08-10
Add ProjectGuid support for SQL projects and CLI

### LINKED ISSUE #965: Consider adding a ProjectGuid to templates and .dacpac
See discussion here:

https://github.com/microsoft/DacFx/issues/824

ErikEJ:
Added --projectguid/-pg CLI option and MSBuild property to set ProjectGuid for SQL projects.

Updated templates and .csproj files to generate and substitute ProjectGuid.

Modified Sdk.targets to pass ProjectGuid to DacpacTool.

Introduced DacOriginModifier to set ProjectGuid in Origin.xml using reflection, integrated into build flow with validation and error handling.

Added unit tests for DacOriginModifier covering correct writing, error handling, and default behavior.

fixes #965


src/DacpacTool/DacOriginModifier.cs
```diff
@@ -0,0 +1,109 @@
...
+                ?? throw new InvalidOperationException($"Unable to locate the Write(Stream) method on {DacOriginTypeName}.");
+
+            var projectGuidProperty = dacOriginType.GetProperty(
+                "ProjectGuid",
+                BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                ?? throw new InvalidOperationException($"Unable to locate the ProjectGuid property on {DacOriginTypeName}.");
+
+            _console.WriteLine($"Setting ProjectGuid {projectGuid} on {OriginPartUri} in {dacpac.FullName}");
+
+            using var package = Package.Open(dacpac.FullName, FileMode.Open, FileAccess.ReadWrite);
```
jmezach: @ErikEJ This now re-opens the Package after it has been written earlier in the build process. Wouldn't it make sense to do this in one go as part of the build?

ErikEJ: @jmezach I think we have to let the internal package build create the Origin.xml before we can add our projectGuid. But maybe I am wrong or misundersand your comment?

jmezach: Hmm, looks like [this line](https://github.com/rr-wfm/MSBuild.Sdk.SqlProj/blob/27d537f28519336063559d0c2d01f7c831157a61/src/DacpacTool/PackageBuilder.cs#L184) is the first time a Dacpac is actually created and until then we're only dealing with a TSqlModel so I guess the Origin.xml doesn't exist yet until that point. That being said, the AddPreDeployment and AddPostDeploymentScript calls just before this is being called also re-open the Package and make changes to it. I wonder if constantly re-opening and saving the file is slowing down builds and whether it could benefit from opening the package just once, making the necessary changes (adding scripts, modifying Origin.xml, and whatever else we might add in the future) could make builds faster.

ErikEJ: @jmezach Should I try to combine all three operations in a single open of the package?

jmezach: @ErikEJ I think that might be worth it yes

jmezach (approved):

ErikEJ:
@jeffrosenberg could you have a look at this?

ErikEJ:
@jeffrosenberg any chance you could look at this?

/cc @jmezach

jeffrosenberg (approved):
Apologies for the delay, seems I missed this the first time!

ErikEJ:
@jeffrosenberg Thanks, no worries!

